PR #8, opened 2015-01-27
produce a moment.js 2.9.0 build

pandeiro:
This is working for me within an adapted version of @martinklepsch 's testbed. I've updated the externs file to include moment.locale(...) introduced in 2.8.1.


martinklepsch:
:beers:

PR #9, opened 2015-01-27
add pt-BR locale for momentjs

pandeiro:


martinklepsch:
@pandeiro I just released `[boot-cljsjs "0.4.4"]` which provides a `minify` task that can be used like so:
```clojure
(minify :in "cljsjs/development/file.js" :out "cljsjs/production/file.min.js")
```
It'd be great if you could integrate that and ideally use the `download` task to download the locale file.

Deraen:
Download is already downloading and extract all the files. Package task would just need to be updated to move locale files to proper location.

We could create deps.cljs entry for each locale?
Then I think it would be possible to (:require [cljsjs.moment.locale.fi-FI]) before calling moment.locale.

pandeiro:
@Deraen This looks good to me, but will break with momentjs >=2.8.1 b/c 'lang' has been renamed to locale, both in the project directory structure and the API. I think this would be an opportune time to bump to 2.9.0 (current stable version) -- what do you think?

PR #24, opened 2015-02-04
Openlayers v3 package

olivergeorge:
This works in my simple (mies based) test.

As discussed, extern is manually from an upstream script.  Noted ticket which may address this.

One odd thing, was an error about the license needing to be (kw str) but I can't see other packages doing that so I've copied their lead.


ol3/README.md
```diff
@@ -0,0 +1,19 @@
+# cljsjs/ol3
+
+[](dependency)
+```clojure
+[cljsjs/ol3 "3.1.1-1"] ;; latest release
```
martinklepsch: `-0` for initial releases please.

ol3/build.boot
```diff
@@ -0,0 +1,30 @@
...
+; NOTE: The externs was generated manually using a script in the upstream repo.  
+; We hope this file may be included in future upstream releases.  A ticket has been
+; raised in relation to this: https://github.com/openlayers/ol3/issues/3202
+
+(deftask package []
+  (comp
+    (download :url "https://github.com/openlayers/ol3/releases/download/v3.1.1/v3.1.1.zip"
+              :checksum "B0C8AEFACA19505A71C8E5002A2E5E26"
+              :unzip true)
+    (sift :move {#"v([\d\.]*)/build/ol.js" "cljsjs/development/ol3.inc.js"})
```
martinklepsch: Since this is essentially the minified version I wonder if it should reside in `cljsjs/production/` instead? (If it doesn't end in `.min.inc.js` it will be supplied to the `:foreign-libs [{:file ..}]` key anyways.) 

Not sure about this, both cases are kind of confusing.

martinklepsch: In the Firebase jar I also put it into development so let's just leave it like that.

martinklepsch:
I wonder if it'd be nice to name this `cljsjs/openlayers` to be a bit more descriptive? The version is reflected by the jar version already so...

ol3/README.md
olivergeorge: Done

olivergeorge:
Good feedback.  I think openlayers is a better package name.  Version tweaked too.

martinklepsch:
I squashed your commits and added them to master. Deployed the jar to Clojars.

martinklepsch:
@olivergeorge I just noticed that open layers provides some CSS as well: http://openlayers.org/en/v3.0.0/doc/quickstart.html 

Is that an important component of OpenLayer that should be included?

olivergeorge:
Yes, I think that's a good idea.

PR #30, opened 2015-02-05
add chancejs package

shaunlebron:
I had to leave out chance.min.js.  It has to be generated with `grunt uglify`.  It cannot be retrieved from the github releases or the npm registry.

### DIFF
--- a/chance/build.boot
+++ b/chance/build.boot
@@ -1,0 +1,26 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[adzerk/bootlaces   "0.1.9" :scope "test"]
+                  [cljsjs/boot-cljsjs "0.4.3" :scope "test"]])
+
+(require '[adzerk.bootlaces :refer :all]
+         '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +version+ "0.7.3-0")
+(bootlaces! +version+)
+
+(task-options!
+ pom  {:project     'cljsjs/chance
+       :version     +version+
+       :description "Chance.js packaged up with Google Closure externs"
+       :url         "http://chancejs.com/"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+    (download :url "https://github.com/victorquinn/chancejs/archive/0.7.3.zip"
+              :checksum "fc430e44258756648d27355ce36180a8"
+              :unzip true)
+    (sift :move {#"^chancejs-\d+\.\d+\.\d+/chance\.js" "cljsjs/common/chance.inc.js"})
+    (sift :include #{#"^cljsjs/" #"^deps.cljs$"})))
--- a/chance/build.boot
+++ b/chance/build.boot
@@ -19,8 +19,11 @@
 
 (deftask package []
   (comp
-    (download :url "https://github.com/victorquinn/chancejs/archive/0.7.3.zip"
-              :checksum "fc430e44258756648d27355ce36180a8"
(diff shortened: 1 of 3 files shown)


shaunlebron:
I ran `boot package build-jar`.  And I tested it with this simple repo:
https://github.com/shaunlebron/cljsjs-chance-test

Also, any ideas on how to retrieve the chance.min.js in the boot script?

martinklepsch:
@shaunlebron there's a `minify` task that can be used like this:
```
(minify :in "cljsjs/development/chance.inc.js" :out "cljsjs/production/chance.min.inc.js")
```
Alternatively you could just use the `download` task a second time and then move the downloaded file with `sift`.

shaunlebron:
great, I will try later today and update this PR

martinklepsch:
Oh and you can use the `deps-cljs` task at the end:
```
(deps-cljs :name "cljsjs.chance)
```
This will automatically create a proper `deps.cljs` based on file extensions of the files in the fileset..

shaunlebron:
There is a trailing comma error that halts the minify task:

```
$ boot package
Downloading 0.7.3.zip
Extracting 24 files
Sifting output files...
Minifying chance.inc.js
Feb 06, 2015 12:00:25 AM com.google.javascript.jscomp.LoggerErrorManager println
SEVERE: /Users/Swilliams/.boot/tmp/Users/Swilliams/code/cljsjs-packages/chance/1ps2/11lgcg/cljsjs/common/chance.inc.js:863: ERROR - Parse error. IE8 (and below) will parse trailing commas in array and object literals incorrectly. If you are targeting newer versions of JS, set the appropriate language_in option.
                        '09' + self.string({ pool: '0123456789', length: 8}),
```

I did some digging, and the `asset-minifier.core/minify-js` doesn't allow any for any of the [usual methods](http://stackoverflow.com/a/22916246/142317) for fixing this.  Will create issue

martinklepsch:
@shaunlebron Ugh. I guess it's easiest/quickest to use the download task then?

shaunlebron:
@martinklepsch there is nowhere to download the minified file.  it must be produced with `grunt uglify`.

martinklepsch:
@shaunlebron http://chancejs.com/chance.min.js

shaunlebron:
Oops, I meant a versioned minified file.  That is a nondeterministic link always pointing to the latest version.

I'll just submit a PR to chancejs gh-pages branch to host version-stamped minified files.  I think that's the correct way forward here.

martinklepsch:
@shaunlebron Ah I see, right. Sorry for my thoughless comment then, hope I didn't come off as a jerk :smiley:

shaunlebron:
ha, not at all.  looks like we have one last problem. The min file ends with a comment referencing the sourcemaps:

```js
//# sourceMappingURL=chance.min.js.map
```

But there's no newline on this last line, which breaks the subsequent concatenated js:

```js
//# sourceMappingURL=chance.min.js.map;(function(){
var k,t=this;
...
```

will investigate where this needs to be fixed

shaunlebron:
submitted [CLJS-1022](http://dev.clojure.org/jira/browse/CLJS-1022) to fix this.  will update when merged

shaunlebron:
[Patched](https://github.com/clojure/clojurescript/commit/0657b0a3ab2a606bb114620d0f66265158a7bc39).  This PR should be safe to merge once a release is cut. will update when that happens

martinklepsch:
Sweet! 👌🎯

shaunlebron:
0.0-2814 released with patch.  Tested it in project repo here:
https://github.com/shaunlebron/cljsjs-chance-test

Good to go!

martinklepsch:
Perfect, thanks!

PR #31, opened 2015-02-05
OpenLayers as native Closure library

martinklepsch:
@olivergeorge @jeluard Please let me know if you think the README could be improved somehow.

### DIFF
--- a/openlayers/README.md
+++ b/openlayers/README.md
@@ -2,17 +2,18 @@
 
 [](dependency)
 ```clojure
-[cljsjs/openlayers "3.1.1-0"] ;; latest release
+[cljsjs/openlayers "3.1.1-1"] ;; latest release
 ```
 [](/dependency)
 
-This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
-of the Clojurescript compiler. After adding the above dependency to your project
-to can require the packaged library like so:
+This jar comes with `deps.cljs` providing OpenLayers **as a Closure library**.
+This means requiring parts of it is done in a similar fashion to
+requiring parts of the Closure library itself:
 
 ```clojure
 (ns application.core
-  (:require cljsjs.openlayers))
+  (:require [goog.dom :as dom]          ; Closure
+            [ol.animation :as anim]))   ; OpenLayers
 ```
 
 [flibs]: https://github.com/clojure/clojurescript/wiki/Foreign-Dependencies
--- a/openlayers/build.boot
+++ b/openlayers/build.boot
@@ -6,7 +6,7 @@
 (require '[adzerk.bootlaces :refer :all]
          '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +version+ "3.1.1-0")
+(def +version+ "3.1.1-1")
 
 (task-options!
  pom  {:project     'cljsjs/openlayers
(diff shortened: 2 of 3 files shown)


jeluard:
That looks good! I will probably add some details about providing configuration options to reduce generated JS size but I need to find some time to experiment first.

martinklepsch:
> configuration options to reduce generated JS size

What do you have in mind there? I'm thinking `:advanced` is all you can get.

jeluard:
OL3 has a number of [defines](https://github.com/openlayers/ol3/blob/master/src/ol/ol.js) you can set to disable some features (like webgl renderer). Those can be provided to the ClojureScript compiler via [:closure-defines](https://github.com/clojure/clojurescript/wiki/Compiler-Options#closure-defines).

Apparently the benefit is pretty big.

martinklepsch:
Interesting, didn't know about this. Curious to see what the effects are :)

olivergeorge:
Wow, a lot happened while I slept.  Well done guys.

I'll run a few tests today.

olivergeorge:
Well my updated example works.  File size is reduced to 240k as expected.  Looks like you've nailed it.

https://github.com/condense/example_openlayers_cljsjs/tree/master/hellool

Yes, the CSS file would be good.  Wasn't sure if there was a system for static assets yet.

olivergeorge:
@jeluard look forward to seeing how :clojure-defines work

jeluard:
I went down to 200k with the following configuration added to `:compiler`:
```clojure
:closure-defines {:goog.DEBUG false
                  :ol.ENABLE_DOM false
                  :ol.ENABLE_VECTOR false
                  :ol.ENABLE_PROJ4JS false
                  :ol.ENABLE_WEBGL false}}
```

I was expecting better results but at least it's something!

olivergeorge:
Nice to know the knobs are there.  Did you put that in your project.clj or
the package deps.clj?

jeluard:
In `project.clj` in the same map where `:optimizations :advanced` is located.

olivergeorge:
I think your tweak was a ~28% filesize improvement...

* 99k    Hello world
* 240k   OL with getting started example
* 200k   OL with getting started example with things disabled

jeluard:
Just noticed OL3 v. `3.2.0` has been released!

Another thing is the huge number of warnings generated during compilation. It is due to some non regular JSDoc annotations being used.
Closure compiler can be configured to accept [extra annotations](http://javadoc.closure-compiler.googlecode.com/git/com/google/javascript/jscomp/CompilerOptions.html#setExtraAnnotationNames(java.lang.Iterable)). That's not yet configurable at the ClojureScript compiler level but could easily be added.

Now the tricky part is that it should be configured at the library level so that end user don't have to mess with that. Is there some mechanism that would allow those to be defined at the library level, merged then used by the ClojureScript compiler?

martinklepsch:
I don't think there's a mechanism for that. Users should pass 
```
:closure-warnings {:non-standard-jsdoc :off}}
```
to the Clojurescript compiler if they want to hide them.

jeluard:
Right but then you miss all potential errors. Something like `:closure-extra-annotations #{"api"}` would be an improvement.
Still it forces all library consumers to add this configuration which is a little sad.

martinklepsch:
> Right but then you miss all potential errors.

In that special case I don't think this can lead to actual errors(?)

In any case packages should under no circumstances be able to modify your build configuration.
So the solution to this is that the packages README documents useful flags that can be used to make the consumption "smoother".

jeluard:
> In that special case I don't think this can lead to actual errors(?)

It will hide all potential warnings coming from others libraries as this is a global configuration. So it will not introduce errors but might hide some potential ones. Otherwise this config should have `:off` as default value?

> In any case packages should under no circumstances be able to modify your build configuration.

I agree with you for the general case but for some particular config like this one I believe it would make sense. Now it makes this feature pretty complex.

martinklepsch:
> It will hide all potential warnings coming from others libraries as this is a global configuration. So it will not introduce errors but might hide some potential ones. Otherwise this config should have :off as default value?

Only the ones that are about JSDoc tags, I doubt those can be fatal. I assume these warnings are mostly useful for companies like Google to catch typos and make sure code follows some conventions.

jeluard:
Probably. In which case it would make sense to make those warnings off by default. I created [CLJS-1015 ](http://dev.clojure.org/jira/browse/CLJS-1015) to track this point.

martinklepsch:
:+1:

olivergeorge:
Nice work!

martinklepsch:
Rebased, merged and deployed as `[cljsjs/openlayers "3.1.1-1"]`.

PR #32, opened 2015-02-09
add CSV parser library

whodidthis:
I just copied the firebase setup, hit me up with any change requirements.


martinklepsch:
Thanks! Two things:

- I think for consistency we should keep the artifacts/provides lowercased (i.e. `cljsjs/CSV` becomes `cljsjs/csv` & `cljsjs.CSV` becomes `cljsjs.csv`).
- If possible it'd be nice for all packages to have minified versions. In the case of firebase this wasn't possible as it defines some `goog.*` stuff but here that's probably not the case. Take a look hat the [hammer build.boot](https://github.com/cljsjs/packages/blob/master/hammer/build.boot) for an example.

There's also a minify task if no minified version is provided.

:100:

martinklepsch:
👌 merging.

PR #34, opened 2015-02-10
google maps extern

rwillig:


martinklepsch:
Thanks @rwillig. Before we merge this please add a Readme similar to how it is done in other package dirs and note that this package is "externs" only in said Readme

rwillig:
got it. will do...

google-maps/build.boot
```diff
@@ -0,0 +1,18 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[adzerk/bootlaces   "0.1.9" :scope "test"]
+                  [cljsjs/boot-cljsjs "0.4.6" :scope "test"]])
+
+(require '[adzerk.bootlaces :refer :all]
+         '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +version+ "3.18")
```
martinklepsch: please add `-0`.

google-maps/README.md
```diff
@@ -0,0 +1,16 @@
...
+```clojure
+[cljsjs/google-maps "3.18-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler.
+
+This library provides an extern file required for compilation in advanced
+mode.
```
martinklepsch: I wonder if this could be made clearer. I.e. people should not expect this thing to do anything, they are not supposed to `: require` it and they'll need an extra script tag to make use of Google Maps JS. I think since most Jars do this stuff it's important here to highlight the differences to avoid confusion. :smile_cat:

rwillig:
adjusted the README to (I hope) make it clear how this jar is different. Let me know if you think it's clear enough

martinklepsch:
:shipit:

martinklepsch:
deployed :+1:

PR #40, opened 2015-02-17
added missing TransitionGroup extern to react-with-addons

levand:
Just a missing extern


martinklepsch:
Thanks! Deployed as `0.12.2-5`.

PR #53, opened 2015-02-22
add mustache.js

piranha:


martinklepsch:
Looking good. Putting `mustache.js` through the [externs generator](http://www.dotnetwise.com/Code/Externs/) I got the following:

```js
var Mustache = {
    "name": {},
    "version": {},
    "tags": {
        "0": {},
        "1": {}
    },
    "clearCache": function () {},
    "parse": function () {},
    "render": function () {},
    "to_html": function () {},
    "escape": function () {},
    "Scanner": function () {},
    "Context": function () {},
    "Writer": function () {}
}
```
Maybe not all of this is needed but it'd be nice if externs are as complete as possible in general.

piranha:
Is it any better now? :)

martinklepsch:
looks good!

PR #55, opened 2015-02-24
Flot charting library

shaunlebron:
[Flot](http://www.flotcharts.org/), a nice charting library:

- contains the 10 standard plugins (used moment locales as reference for packaging)
- included an example/ for trying it out


martinklepsch:
Hey. Sorry for taking so long to respond. Two remarks/questions:

- Are you sure the externs are enough especially given the list of plugins that, I assume, adds more functions?
- I think I'd like to keep "example" like stuff out of this repo for now. I see that these kinds of examples are useful for testing though so there probably should be a place where we can collect them. (I created the [testbed](https://github.com/cljsjs/testbed) repo when I came to that conclusion.)

If you have any thoughts or suggestions on the second point they're very welcome!

shaunlebron:
I went back and checked, and you're right, there are some plugins add public methods that will require externs.  Thanks, will add those and remove the example.

shaunlebron:
Made the changes.  Would you prefer a squashed PR?  I meant to do that, but somehow messed it up.

martinklepsch:
Squashed would be nice but isn't required for merging. The example seems to be still present looking at the PR diff? I tried to squash locally but also messed it up. Maybe easiest thing to do is to create a new branch copy the relevant files & open new PR? :smile:

PR #59, opened 2015-02-28
add flot library

shaunlebron:
squashed the commits from #55 (example is removed)

### DIFF
--- a/flot/README.md
+++ b/flot/README.md
@@ -1,0 +1,34 @@
+# cljsjs/flot
+
+[](dependency)
+```clojure
+[cljsjs/flot "0.8.3-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.flot))
+```
+
+You can also require individual packaged plugins:
+
+```
+cljsjs.flot.plugins.canvas
+cljsjs.flot.plugins.categories
+cljsjs.flot.plugins.crosshair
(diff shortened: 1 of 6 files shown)


martinklepsch:
Great. I see you added more externs, are those complete? (Some seem to relate to the different plugins.)

I think it's ok if we add externs in a `best-effort` style as there is really no way of verifying/testing them so if you think they're good as is, let me know and I'll merge.

shaunlebron:
Externs are complete.  Most plugin features are internally driven by additional input options for the original flot constructor, so most don't provide any public methods on the returned plot object.  I included the externs for those that do.  Good to go.

martinklepsch:
Merged. Will do some minor adjustments tomorrow and leave a note here when I pushed it to Clojars :)

Thank you!

martinklepsch:
Pushed `cljsjs/flot-0.8.3-0`. I wonder why you decided to put some plugins into `common` while others are in `plugins`?

shaunlebron:
Thanks for reviewing. The plugin externs are in `cljsjs/common/`, but the plugin code is in `cljsjs/plugins/`.  I was kind of loose on that, let me know if you want changes.

martinklepsch:
I think since the files in `plugins` could be split up perfectly into `production` and `development` it would be an appropriate change. 

Also if you decide to make the change feel free to nest the `production`, `development` & `common` dirs in a directory named after the package as described [in the wiki](https://github.com/cljsjs/packages/wiki/Creating-Packages).

I'll add you to the CLJSJS org so you can push these kinds of changes as well. 
If you'd like to be able to deploy to Clojars as well, I'd be happy to add you there as well :)

shaunlebron:
Great, thanks.  will check in those changes.  Will ping you later for clojars access, thanks.

PR #61, opened 2015-03-01
Highlight.js - syntax-highlighting

shaunlebron:
[Highlight.js](https://highlightjs.org/) is a syntax-highlighting library. Its website allows you to select which languages you want to highlight, and it creates a packaged javascript for you to download.

I had a [discussion](https://github.com/isagalaev/highlight.js/issues/761) with its contributors about packaging it for cljsjs.  Looks like their cdn build configuration produces all the files that we need, so that we can use the library like this:

```clj
(ns example
  (:require
    cljsjs.highlight
    cljsjs.highlight.langs.clojure))
```

Since the highlight.js library files must be built from [source](https://github.com/isagalaev/highlight.js), I included them in the resources directory.  They were created by running the following in highlight.js repo:

```
$ git checkout 8.4
$ node tools/build.js -t cdn :none
$ cp -r build/* <path-to-cljsjs-packages>/highlight/resources/cljsjs/common/
```


shaunlebron:
sorry I'm rushing out the door, but I added the wrong js files from master branch, not the latest release 8.4.  I'll fix this later.

shaunlebron:
alright, force-pushed the updates

martinklepsch:
Can we use the `download` task here so we don't need to store all source files in the repository?

shaunlebron:
The files aren't hosted anywhere.  If we used the `download` task, we would have to run the following to produce the source files:

```
node tools/build.js -t cdn :none
```

Is this possible?

martinklepsch:
Hm. We could integrate something like that into the build process although having them downloaded from somewhere would be nicer. Can't believe they're not distributed with builds...

shaunlebron:
Yeah, this is the second time this has happened. Chance.js didn't have deterministic download links.  And Highlight.js has several build permutations, but only the common one is published.

I think in the long run, we need a way to shell-out to the node/grunt build to account for these cases.

martinklepsch:
@shaunlebron yeah, I suspect this is something that has to be worked out. There is `boot.util/dosh` which we can use for that kind of stuff. Maybe we can add a task to the cljsjs packaging tasks that tests if some command is present and runs some stuff if it is. 

Are you fine with putting this PR on hold until we implemented something like the above?

martinklepsch:
@cljsjs/contributors thoughts?

Deraen:
Great work! It's nice to see that the languages can be distributed as a separate namespaces, last I checked Highlight.js some months ago I got really frustrated and just used the website to generate a working build.

About the files, yeah it would be nice to download the files as a part of the build process. I guess it should be possible to spawn a node process from Boot, only problem is to use proper temp directory as the working directory for the node process. And that shouldn't be a problem as all the files a downloaded to one directory?

pandeiro:
@shaunlebron : Just checking, you looked at [the release repo](https://github.com/highlightjs/cdn-release), right? That does not meet our needs here?

shaunlebron:
@pandeiro I'm embarrassed I didn't find that :/ We can use the language files there.  Last issue is `highlight.min.js` there being packaged with common languages.  The build process I was using strips those out so we can opt-in to the languages we want.  If we're fine with that, we can just use it.

pandeiro:
@shaunlebron Ha! Nothing to be embarrassed about -- I found that trawling its Open Issues to see if anyone else had complained about the lack of a readymade build. Not exactly easy to find.

About stripping the common stuff -- sounds very cool but I think for cljsjs it would make sense to deliver the vanilla, w/out any surprises. What do you think?

shaunlebron:
@pandeiro Ha, alright.  I like the no-surprises approach.  I'm not sure how much people care about file sizes, but the core is 7kb stripped, and 37kb with the common languages.

pandeiro:
Yeah that is a big difference. Maybe it could be delivered a la react/react-with-addons as two packages: highlight/highlight-bare? How do you do the stripping?

shaunlebron:
Agreed.  We can strip the core either by running:

```
node tools/build.js -t cdn :none
```

or we can process the distributed highlight.min.js with a simple regex replace:

```clj
(replace s #"(?s),hljs\.registerLanguage.*" ";")
```

martinklepsch:
Maybe we can take the core file from one repo and the language files from
another?

(Just guessing here, might be overlooking sth.)

On Wednesday, March 11, 2015, Shaun LeBron <notifications@github.com> wrote:

> Agreed. We can strip the core either by running:
>
> node tools/build.js -t cdn :none
>
> or we can process the distributed highlight.min.js with a simple regex
> replace:
>
> (replace s #"(?s),hljs\.registerLanguage.*" ";")
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cljsjs/packages/pull/61#issuecomment-78377574>.
>


-- 
—
Martin
http://martinklepsch.org <http://martinklepsch.org?ref=email>

shaunlebron:
I don't know if the stripped/minified core is hosted anywhere.  And we can't use our `minify` task because their minifier manually renames symbols shared with language files, which we are now pulling from the cdn repo.

I'll ask if they'll host the stripped file in their cdn repo.

Deraen:
I implemented a version which downloads the sources and builds the files using tools/build.js.

https://github.com/cljsjs/packages/tree/highlight

Implementation is not the prettiest but we can fix this later by moving the task to boot-cljsjs. I did a quick test using the package and this seems to work fine.

Does anyone have thoughts how to name the language and style files? e.g. `cljsjs/common/styles/tomorrow-night.css` doesn't have any info about which package this file belongs to so there is a change of classpath coflicts.

Deraen:
Would having language and style files under `cljsjs/common/hightlight/` sound reasonable?

Deraen:
I decided `cljsjs/common/hightlight/` is a good folder for language and styles files and merged the changes and deployed the package to clojars!

pandeiro:
@Deraen: :+1: on `common/<package-name` being the convention for all package assets

martinklepsch:
@Deraen @pandeiro I recently followed the following convention:
```
cljsjs/highlight/common
cljsjs/highlight/development
cljsjs/highlight/production
```

Are you suggesting to keep the `cljsjs/<pkg-name>/{development,production}` but put other assets into a `cljsjs/common/<pkg-name>` directory? I'm thinking it's probably least confusing if we follow the same pattern for all three directories.

pandeiro:
No, @martinklepsch, I was confused. I think you're right that cljsjs/<package>/etc is best, as it mirrors the artifact most closely. This is already specified as such in the [wiki article](https://github.com/cljsjs/packages/wiki/Creating-Packages), so I think we're good to go.

Deraen:
Oh, I was unaware of that convention. Seems a good choice.

So we should fix at least highlight package, and I think there are several others also that don't follow that. Maybe the libraries with no resources in addition to JS sources (e.g. react) should still follow cljsjs/common/{package}.ext.js.

pandeiro:
@Deraen I was thinking about that too but it entails a cljsjs version bump to a majority of the packages. Is this a good idea? One alternative might be to wait for actual version bumps / PRs and make this change then, when necessary. That has the downside though of us forgetting to do it.

Re: react, I think this only really matters when users need to access the additional resources, so in the case that none exists, I don't think it matters -- is it possible for externs files to conflict? (If two JS libs have the same name?)

martinklepsch:
@pandeiro I don't see much of an issue with version bumps. Any CLJSJS increment should be non-breaking and safe to update anyways.

> Maybe the libraries with no resources in addition to JS sources (e.g. react) should still follow cljsjs/common/{package}.ext.js.

Why not `cljsjs/react/common/react.ext.js`? This way there's almost no chance of conflict and the different "profiles" (common, production, development) would all be in the same spot. Is your thinking that users usually don't want (to serve) the .ext.js file? While it's probably the case I don't see an issue with putting it in the same place as CSS etc.

Deraen:
My only concern is that moving files might break stuff.

pandeiro:
@martinklepsch Yeah I figured `.ext.js` exist for the cljs/gclosure compiler, not applications, and we have a lot of libs already. Who's gonna do the heavy lifting here? ;)

martinklepsch:
All files listed in `deps.cljs` can be moved safely as far as I understand. 
What might break is CSS import helpers people wrote to get CSS out of
CLJSJS packages.

olivergeorge:
Are there commonly used CSS import helpers?

martinklepsch:
@olivergeorge There's a wiki page about it but no "definitive guide" yet.

PR #63, opened 2015-03-04
Update google-maps to 3.18

paraseba:
The README file states that it is 3.18, but the actual externs were for 3.15.
3.18 includes many new features that are now covered by this externs file.

The externs file was taken from https://github.com/google/closure-compiler/blob/master/contrib/externs/maps/google_maps_api_v3_18.js


martinklepsch:
Thanks Sebastian, good catch :100: 

I pushed a new version with the updated externs.

PR #71, opened 2015-03-21
Add highlight package

mynomoto:


Deraen:
Hi,

There is already a another PR about Highlight.js #61.

mynomoto:
Sorry, should have checked. That one is way better than this. Just wish that it would be merged already.

PR #73, opened 2015-03-21
Fixed data table

ul:
Add [FixedDataTable](https://github.com/facebook/fixed-data-table) React component and ES6 `Object.assign` polyfill required by it.


olivergeorge:
Thanks @ul 

@martinklepsch could you please check you're happy with our approach.  In particular, fixed-data-tables uses and es6 feature (Object.assign) so we've packaged it separately. 

There's a demo here if you're curious to see how it can be used:
https://github.com/ul/fixed-data-table-demo

Deraen:
Object.assign from https://github.com/es-shims/es6-shim is available as standalone file from  https://github.com/ljharb/object.assign. Unfortunately it looks like it would require CommonJS support.

Though I'm not completely happy with having implementation here, I think it acceptable in this case.

olivergeorge:
Thanks @Deraen 

I thin there's reason to hope that it's only a temporary patch.  The google closure wiki talks of [transpilation of ES6](https://github.com/google/closure-compiler/wiki/ECMAScript6) and there are associated [language-in / language-out flags](https://github.com/clojure/clojurescript/blob/master/src/clj/cljs/closure.clj#L104) in the clojurescript compiler.  Having said that, I didn't come across direct discussion of things like Object.assign.

olivergeorge:
Just chatting with David.  Sounds like you can set language-in to es6.  By rights that might mean we don't need this shim.  #needstesting

ul:
Thanks, it is interesting idea I wish it will help. Will test it tomorrow.

ul:
Does not work for me. Not sure if it's because I misused language-in option, or it does not help in that case. I set `:language-in "ecmascript6"` in test project and required FDT version without a shim, and got the same `Object.assign is not a function` error.

olivergeorge:
Hi Ruslan, thanks for trying.

PR #78, opened 2015-03-24
[document-register-element] Added 0.2.1

jeluard:
Introduced the cljsjs package for [document-register-element](https://github.com/WebReflection/document-register-element).

### DIFF
--- a/document-register-element/README.md
+++ b/document-register-element/README.md
@@ -1,0 +1,21 @@
+# cljsjs/document-register-element
+
+[](dependency)
+```clojure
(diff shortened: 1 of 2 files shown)


jeluard:
What do you guys think about this one? Can I go ahead and merge it?

martinklepsch:
I think it'd be good to have a note in the read me why there are no
externs, other than that looks good :)

jeluard:
Great!

martinklepsch:
Awesome, thank you! :)

PR #81, opened 2015-03-27
Update README.md

konrad-garus:
It seems that the -1 version in Clojars is still broken, and -2 is the correct one.


martinklepsch:
Thanks :+1:

PR #83, opened 2015-03-28
[heap] Added 1.0.0

jeluard:
Heap JS currently doesn't define a version number. I discussed with heap guys and then will consider it in the future.
For now I just used `1.0.0`. As you can't choose your JS file but have to use whatever their latest is I don't think it is a problem.

### DIFF
--- a/heap/README.md
+++ b/heap/README.md
@@ -1,0 +1,14 @@
+# cljsjs/heap
+[](dependency)
+```clojure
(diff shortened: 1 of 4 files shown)


martinklepsch:
What'd of you think about using something like `0.0.1`? If they decide to release a versioned file with a version below 1.0.0 the package version might be confusing.

jeluard:
Good point! I just made the change.

martinklepsch:
:100:

PR #87, opened 2015-04-06
[waypoints] Added 3.1.1

RadicalZephyr:
Add support for the Waypoints library.


martinklepsch:
Pushed this as `cljsjs/waypoints "3.1.1-0"]`.
We usually append `-0` for initial releases so we can release new versions with updated externs or similar fixes without messing with the version of the packaged library.

Thanks for the PR! :beers:

PR #96, opened 2015-04-13
Change build.boot to download and sift raw JS instead of zip so that loc...

hipitihop:
Change build.boot to download and sift raw JS instead of zip so that local build works on Windows/Linux

Note: on Windows 7 previously `boot package build-jar` complains about not being able to delete the downloaded zip, and fails when sift regex has forward slashes in path. Using raw JS avoids this problem. I suspect other projects here may have similar issues under Windows.

### DIFF
--- a/pako/build.boot
+++ b/pako/build.boot
@@ -19,10 +19,11 @@
 
 (deftask package []
   (comp
-    (download :url "https://github.com/nodeca/pako/archive/0.2.6.zip"
-              :checksum "dba1e8029dc001e640a4f1dc034ca7c2"
-              :unzip true)
-    (sift :move {#"^pako-.*/dist/pako.js"    "cljsjs/pako/development/pako.inc.js"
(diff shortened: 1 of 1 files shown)


martinklepsch:
@hipitihop I'd rather fix this "once-and-for-all" by changing `boot-cljsjs` to work smoothly on Windows. Would you be up for giving that a whirl? :smiley:

hipitihop:
Better windows support is welcome. Happy to help to be a guinea pig. I'm no Windows dev anymore, Ubuntu/Linux here with access to Win 7 VM, others on our team have Win so can also test.

martinklepsch:
I added an issue to the `boot-cljsjs` project and will close this PR for now.

martinklepsch:
@hipitihop please subscribe to the linked issue and let people know that you can try out things if that's still the case :)

PR #97, opened 2015-04-14
Adding codemirror 5.1.0 cljsjs package

jmatsushita:
Hi there,

This works for me, hopefully it'll be useful for others!

Jun


jmatsushita:
FYI, this doesn't have modes yet.

martinklepsch:
@jmatsushita thanks, this looks really good! You also committed the js and css files which shouldn't be required as those are fetched during `boot package`. If you can remove those that would make it perfect!  :+1:

jmatsushita:
Hi @martinklepsch, thanks for the input! I got the modes in as well, so now it's doing something useful :)

martinklepsch:
@jmatsushita wow, this turned into a truly amazing PR! :star2: 

I wonder if we should minify the files for the separate modes? Any opinions? /ping @Deraen

martinklepsch:
@jmatsushita It's too good to keep it back. I just pushed it to Clojars and merged your changes! Thanks a ton!

PR #98, opened 2015-04-17
Fix Showdown under advanced compilation

bobby:
The .makeHtml function was getting swallowed by advanced compilation, so I fixed the externs file.


bobby:
Whoops, pulled in some previous commits to css-layout.  I'll isolate the Showdown changes.

PR #92, opened 2015-04-09
[gl-matrix] Add gl-matrix library.

jenanwise:
Currently using my own fork of the library because the original hasn't had a proper release in a while. My fork is just a build of their current master, which includes some crucial bug fixes. I wasn't sure how to semver this so I used `MAJOR.MINOR.RELEASE-FORKNAME-0`: `2.3.0-jenanwise-0`.

* Original library: https://github.com/toji/gl-matrix
* Fork: https://github.com/jenanwise/gl-matrix

Test project verifying `:none` and `:advanced` both work: https://github.com/jenanwise/cljsjs-gl-matrix-test

PR #101, opened 2015-04-19
[gl-matrix] Use tag url instead of master for downloads.

jenanwise:
I forgot to use the proper tag the download urls in the original PR #92!


martinklepsch:
:+1: Thanks!

PR #102, opened 2015-04-19
add react-bootstrap

tscheibl:


tscheibl:
I forgot to include the Bootstrap dependency and instructions on how to use this with Boot (boot-less, ...). I will submit a new pull request after I fixed this.

PR #107, opened 2015-04-22
update react router to 0.13.2

ghedamat:
sadly regenerating the externs is now harder because react router has moved to UMD builds 

They *should* not have changed but I'll try to do some more investigation

### DIFF
--- a/react-router/build.boot
+++ b/react-router/build.boot
@@ -19,8 +19,8 @@
        :license     {"BSD" "http://opensource.org/licenses/BSD-3-Clause"}})
 
 (deftask download-react-router []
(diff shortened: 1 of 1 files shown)


martinklepsch:
Hey @ghedamat!

> They should not have changed but I'll try to do some more investigation

Have you come up with a conclusion? Ok to merge as is?

ghedamat:
@martinklepsch I wasn't able to find breaking changes in my use of the library
nor new apis from the changelog

so I guess it's good to merge

that said
if anyone knows a way to generate externs from UMD builds I'd love to learn it!

PR #113, opened 2015-05-04
[pouchdb] Fix externs file

fehrenbach:
Whoops

That's what you get for not testing advanced compilation.
It works for me™ now, but it might still be incomplete or otherwise broken.


martinklepsch:
@fehrenbach :100: Deployed to Clojars :)

PR #116, opened 2015-05-07
Matter.js 0.8.0 initial add

bradleybuda:
Adding [matter.js](https://github.com/liabru/matter-js).

I tested with `boot package build-jar` and a local require and it seems to work, but this is my first encounter with cljsjs, and I'm a relative Clojure newbie, so please let me know what can be improved.


martinklepsch:
Closed via f58d4e5

PR #117, opened 2015-05-09
Add markdown-js package

mynomoto:
Sorry about adding files directly. The problem is that we don't have a `tar -zxvf` task yet and I shaved more yaks than I should over the last days :(


Deraen:
`[cljsjs/boot-cljsjs "0.4.8"]` now contains `decompress` task which can be used to decompress tar gz files.

E.g.
```clj
    (download :url "https://github.com/evilstreak/markdown-js/releases/download/v0.6.0-beta1/markdown-browser-0.6.0-beta1.tgz"
              :decompress true
              :compression-format "gz"
              :archive-format "tar")
    (sift :move {#"^markdown-browser-.*/markdown\.js" "cljsjs/development/markdown.inc.js"
                 #"^markdown-browser-.*/markdown\.min\.js" "cljsjs/production/markdown.min.inc.js"})
```

Deraen:
I would like if you squashed the commits into one.

Other than that, looks good!

mynomoto:
Replaced by #122

PR #119, opened 2015-05-12
Klayjs

moorooboorai-jeroen:
CLSJS package for [KlayJS](https://github.com/OpenKieler/klayjs), a layer-based layout algorithm.

### DIFF
--- a/klay/README.md
+++ b/klay/README.md
@@ -1,0 +1,45 @@
+# cljsjs/klay
+
+[](dependency)
+```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
- I think the first `var $klay` in extern is unnecessary?
- Licence should be EPL-1.0 instead of BSD?

Deraen:
And as you mentition in example, using `#js` looks clunky, so perhaps examples should instead use `clj->js`?

PR #124, opened 2015-05-18
Add CodeMirror event functions on/off

magnars:


magnars:
Here's a link to the codemirror docs referencing these functions: https://codemirror.net/doc/manual.html#events

martinklepsch:
@magnars Thanks. I merged this and pushed a new version. :+1:

PR #130, opened 2015-05-27
package chemdoodle web

zachcp:
This is a cljsjs package for the  [ChemDoodle WebComponents](http://web.chemdoodle.com/) library. It excels in depicting molecules in the browser and, along with [kemia](http://kemia.github.io/), it could represent a near-complete browser based solution for common chemical manipulations/depictions.

Note: Chemdoodle also has an excellent "Sketcher" functionality that depends on the core library. I tried packaging this up initially but the libraries must be loaded sequentially (base library first). It looks as if the way to handle this is to package them separately and have the UI depend on the base. If there is a way to package these together, I'm all ears.

Thanks,
zach cp


Deraen:
There shouldn't be need to package sequentially loadable libs separately but you might have to write the deps.cljs by hand, something like this should do:
```clj
{:foreign-deps [{:file "cljsjs/chemdoodle/development/chemdoodle.inc.js"
                 :file-min "cljsjs/chemdoodle/produciton/chemdoodle.inc.min.js"
                 :provides ["chemdoodle"]}
                {:file "..."
                 :file-min "..."
                 :provides ["chemdoodle.sketch"]
                 :requires ["chemdoodle"]}]
 :externs ["..."]}

martinklepsch:
Is this ok to merge?

zachcp:
the core functionality of this library works just fine so I think its okay to merge. I can add the sketching libraries to another pull request. If, however, you would like me to get everything together for a single pull request I am happy to do it that way as well.

PR #135, opened 2015-06-06
Update mui to 0.1.8

mynomoto:


martinklepsch:
Do I understand correctly that externs are not needed for the `mui.js` file because you will never directly call any of the functions in there?

mynomoto:
Yes, that's it.

Deraen:
How are you then using this? Is it only for CSS?

mynomoto:
Loading the library add event listeners for changes on DOM then it changes elements with specific classes.

Deraen:
Okay.

PR #136, opened 2015-06-07
adding the dimple namespace to the dep.cljs file

zachcp:
I was looking through a few of the cljsjs packages while making my own and I noticed that `dimple` differs from the recommended folder hierarchy. The fix is simple: put the dev and prod folders inside of the dimple folder.

target/
├── cljsjs
│   └── <library-name>
│       ├── common
│       │   └── <library-name>.ext.js
│       ├── development
│       │   └── <library-name>.inc.js
│       └── production
│            └── <library-name>.min.inc.js
└── deps.cljs


martinklepsch:
Nicely spotted, thanks!

PR #138, opened 2015-06-11
FixedDataTable version 0.3.0 package

rundis:
Never worked with clsjs (and very little with boot) before, so I'm not sure how to test whether this actually works. The package task seemed to do it's thing though.


martinklepsch:
To install the jar locally you can run `boot package build-jar`. After that you should be able to use it in your projects (it will be installed into your local maven repo). 

A bit of smoke testing would be great before merging this :)

Thank you!

rundis:
Cheers !

Ok I've installed it locally and tried with a basic table. Both in dev mode and with advanced/minified compilation and it seems to work ok. Not an exhaustive test of all datatable props though.

We are planning on using this version in our project so I have a vested interest in ensuring we can use it. 
Let me know if you think I should do more exhaustive testing, otherwise I'm sure I'll be submitting a new pull request if we find any problems (that can be traced to for example the externs file)  !

cheers
Magnus

PR #155, opened 2015-06-29
[hammer.js]: add set/get method; jump version; fix checksum error

liusong1111:


martinklepsch:
Thank you very much, deployed and merged :+1: 
Sorry for taking so long

PR #156, opened 2015-06-29
add fabric.js

liusong1111:
Add fabricjs:
[fabricjs.com]

Fabric's source code is long(2w+ lines),and I spent about one day to figure out the class/method/attribute names manually, and tested in my personal project.  It may be something missing though.


martinklepsch:
Thanks a lot. Those externs really look like a crazy amount of work. :100: 
Sorry for letting this sit for so long.

PR #152, opened 2015-06-23
[react-bootstrap] Update to 0.23.5

dmcgillen:
Updating react-bootstrap from 0.21.2 to 0.23.5, hopefully correctly!

I based this change on the previous commit that had updated the react-bootstrap version, using http://www.dotnetwise.com/Code/Externs/ to generate react-bootstrap.ext.js as noted in the file's comments

PR #160, opened 2015-07-14
Updated react-boostrap to 0.23.7

ricardojmendez:
New archive references and checksum, also added externs for the two new components I saw on their changelog.


martinklepsch:
Seems that #152 contains an update for 0.23.7 as well?

ricardojmendez:
Yes, look like it does.

PR #163, opened 2015-07-17
[classnames] Added classnames v2.1.3 to cljsjs/packages

bensu:
It is a very small package that will be needed for https://github.com/jquense/react-widgets

### DIFF
--- a/classnames/README.md
+++ b/classnames/README.md
@@ -1,0 +1,18 @@
+# cljsjs/classnames
+
+[](dependency)
+```clojure
+[cljsjs/classnames "2.1.3"] ;; latest release
+```
+[](/dependency)
+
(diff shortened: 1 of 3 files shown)


crisptrutski:
Talk about a microlibrary! Had to check usage to convince myself it could even be useful :)

Have released as `2.1.3-0`, note the `-0` build number added to be consistent (though I doubt these externs will change..). (I'm new here and just cargo culting)

Those widgets are so clean and simple, nice find..

crisptrutski:
Came in as 65c8d9efda7b211c682201deb63ed30ca2415ee5

bensu:
Thanks, I actually need this for another library, https://github.com/jquense/react-widgets and I will try to add all of its dependencies to the project. I'm still not sure if it will be possible to properly package it under cljsjs.

I'll have the -0 into account for the next one.

Issue #153, opened 2015-06-24
Papaparse has incorrect name in extern file


erikcw:
I just tried compiling (:advanced mode) a project using `papaparse` and get the following warnings:

```
Jun 24, 2015 1:12:49 PM com.google.javascript.jscomp.LoggerErrorManager println
WARNING: file:/Users/erik/.m2/repository/cljsjs/papaparse/4.1.1-0/papaparse-4.1.1-0.jar!/cljsjs/papaparse/common/papaparse.ext.js:3: WARNING - name papa is not defined in the externs.
papa.parse = function (csvData, config) {};
^

Jun 24, 2015 1:12:49 PM com.google.javascript.jscomp.LoggerErrorManager println
WARNING: file:/Users/erik/.m2/repository/cljsjs/papaparse/4.1.1-0/papaparse-4.1.1-0.jar!/cljsjs/papaparse/common/papaparse.ext.js:4: WARNING - name papa is not defined in the externs.
papa.unparse = function (csvData, config) {};
^
```
https://github.com/cljsjs/packages/blob/master/papaparse/resources/cljsjs/papaparse/common/papaparse.ext.js#L3

```javascript
var Papa = {};
papa.parse = function (csvData, config) {};
papa.unparse = function (csvData, config) {};
```

It should be `Papa.[un]parse` (uppercase "P" in Papa).

Deraen:
Should be fixed now.

PR #164, opened 2015-07-17
[papaparse] Fix extern case #153

snoe:
Fixes advanced compilation errors as detailed in #153


martinklepsch:
Thank you. I published `4.1.1-1` on Clojars! :beers:

PR #165, opened 2015-07-17
Add virtual-dom

exupero:
Adds a package for the JavaScript bits of the [vdom](https://github.com/exupero/vdom) library.

Let me know if anything is amiss.


martinklepsch:
@exupero Thanks for the PR! I'm not sure I fully understand what this is. The description has words like "ClojureScript wrapper" but then it's a Javascript file? Would you care to elaborate a bit more?

crisptrutski:
It appears that vdom.js is a build of [virtual-dom](https://github.com/Matt-Esch/virtual-dom)? Perhaps it makes more sense to wrap `virtual-dom` itself for use in `vdom`.

Perhaps I'm misunderstanding something though - is there more to these `vdom` js files than vendoring?

exupero:
@martinklepsch Good catch. The description needn't say anything about ClojureScript. It's really just a build of virtual-dom with a few specific functions exposed under particular names.

@crisptrutski You're correct. This is just a build of virtual-dom and could be packaged under that name rather than `vdom`. Are any changes necessary to the exposed API to make it a general purpose build of virtual-dom?

crisptrutski:
@exupero thanks for clarifying. Perhaps you wouldn't mind skimming the API yourself? Don't want to be a bottleneck, I might not have time today.

Beyond extending externs (if necessary) and renaming, LGTM :smile:

martinklepsch:
Maybe then this should also be packaged as `cljsjs/virtual-dom`? I still don't quite see what `vdom` is and what differentiates it from `virtual-dom` to be honest.

exupero:
Done. Thanks for the suggestions. Let me know if you see any other room for improvement.

crisptrutski:
Thanks for the update, two questions though:

1. Is it not possible to pull the js from the official repo's dist builds?
2. What do you think about adding the `h` aka "virtual superscript" function to externs? This hiccup-alike seems part of the public API.

exupero:
1. Possibly. The official repo's build does not include `virtual-dom/virtual-hyperscript/svg`, which is a primary use case for me. Neither does the official repo include a minified version of the build. I'm open to other solutions.
2. Will gladly include `h`.

I also see that in the official build's `create-element` is listed in the API as `create`, which I can imitate.

crisptrutski:
In my opinion bundling virtual demo directly is best - easier to track version, and no surprises for use cases other than vdom. 

Minification could be generated as part of the `package` task. Since Closure has dead code elimination, including additional pieces not part of their default build seems OK here, again ideally concatted as part of the `package` task.

@exupero sound reasonable?

@martinklepsch please jump in if you have other preferences

crisptrutski:
@exupero Actually I've been told that DCE does not apply to `:foreign-libs` code in general. Perhaps a better solution for the `virtual-hyperscript/svg`, and any other non-standard modules you require, to have their own packages

exupero:
Sounds reasonable.

I'll use virtual-dom's build directly and minify as part of the `package` task. I'll look into creating a new package for `virtual-hyperscript/svg`.

Thanks for the feedback.

martinklepsch:
The reason for not tracking `virtual-dom` directly seems to be that the file provided is not comprehensive enough, is that correct? Maybe the author of `virtual-dom` could be convinced to also provide a "complete build" of the library?

@exupero sorry for being so hesitant to just merge this. I'd just much rather like to track the original repo than a custom build maintained by a single person. I hope we can find a good solution together here :)

@Deraen you've packaged some libs with submodules and stuff — maybe you have other suggestions?

exupero:
@martinklepsch Correct, it does not include some submodules I commonly use. I'm open to suggestions on how to include them.

Hesitation is not a problem. Being unfamiliar with CLJSJS, I opened the pull request as a starting point, and I appreciate the feedback.

Deraen:
I see this is currently using the official build. Does it still miss some modules?

The dist file is build using browserify. How are the modules selected?

One way would be to unpack full zip to temp-dir and then call necessary programs to generate required files, e.g. virtual-dom.js and virtual-dom.all.js. Like highlight js is built: https://github.com/cljsjs/packages/blob/master/highlight/build.boot#L45-L57

I don't know browserify well enough to know if it's possible to expose modules as separate files, but if it's possible, that would be even better.

exupero:
It's missing a virtual-dom module that I use (`virtual-hyperscript/svg`), but I can work around it and it doesn't need to be part of the CLJSJS package.

The modules are specified in the [main module definition](https://github.com/Matt-Esch/virtual-dom/blob/master/index.js), and Browserify includes all of them in the output file. An alternative build file with more modules could be created, outputting a virtual-dom.all.js file, but that's up to the virtual-dom maintainers.

crisptrutski:
Superseded by #187

PR #169, opened 2015-07-23
[js-yaml] Add library

Elyahou:


crisptrutski:
Thanks! 

Having an issue testing this with advanced compilation - the entire library is getting is getting dead code eliminated. Perhaps I'm doing something silly though - have you tested this on your side?

@martinklepsch 
Side nit is that the library is using `double` (a reserved word up before ES5), the minifier is complaining. AFAIK CLJS does emit ES5 though, perhaps this is the underlying issue: https://github.com/yogthos/asset-minifier/issues/6

crisptrutski:
OK, mispoke about DCE - the `js-yaml.min.inc.js` file produced when I ran `boot package` is empty. This is likely related to the reserved word errors. @Elyahou can you confirm whether you saw these warning during packaging?

Elyahou:
Same behavior here too, Sorry I did'nt tested on advanced before submitting PR...

Just pushed new version with the minified js directly from the package as it seem that minify task have issue with this file.

crisptrutski:
Thanks! Merged and published :tada:

PR #174, opened 2015-08-01
[web-animations-js] Added 2.1.2.

jeluard:
[web-animations-js](https://github.com/web-animations/web-animations-js) does not come with usable non minified js so I just use the `min` for both `inc.js`. Not sure this case is supposed to be addressed.

### DIFF
--- a/web-animations-js/README.md
+++ b/web-animations-js/README.md
@@ -1,0 +1,20 @@
+# cljsjs/web-animations-js
+
+[](dependency)
+```clojure
+[cljsjs/web-animations-js "2.1.2-0"] ;; latest release
+```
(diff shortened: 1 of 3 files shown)


martinklepsch:
@jeluard I'm updating the package name to `web-animations` as we generally omitted any trailing `js` things. Will be up in a bit.

martinklepsch:
Thank you Julien :+1:

martinklepsch:
@jeluard ah btw. I removed the `.gz` and `.map` files. Was it intended to package those or were they more an unintended result of the regex?

jeluard:
Nice catch! Thanks for doing this check!

PR #176, opened 2015-08-04
[auth0-lock] add library @ 7.7.1

tmcf:
Hi,
This pull request adds support for auth0-lock, which provides sign-in dialogue boxes for use with auth0. This is not the same library as auth0, which is already in cljsjs.

I have provided extern definitions for all javascript functions that were documented as public. dev and min builds have been tested in a clojurescript project for the main entry points.

This is my first cljsjs contribution, hopefully I have not overlooked anything major. 

Thanks,
Ted

### DIFF
--- a/auth0-lock/README.md
+++ b/auth0-lock/README.md
@@ -1,0 +1,20 @@
+# cljsjs/auth0 -lock
+
+[](dependency)
+```clojure
+[cljsjs/auth0-lock "7.7.1-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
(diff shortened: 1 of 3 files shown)


auth0-lock/README.md
```diff
@@ -0,0 +1,20 @@
...
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.auth0))
```
martinklepsch: please use `cljsjs.auth0-lock` here as well :)

martinklepsch:
Looking great besides the minor README tweak. Just went ahead and did it. 
Thanks @tmcf, great first contribution! :+1:

Should be on Clojars now :)

tmcf:
Thanks!

PR #178, opened 2015-08-04
Updating old [Foreign Libs] URLs, add missing references

crisptrutski:
Noticed that all the [Foreign Libs] URLs were broken, and in some cases the `[flibs]` alias was not even defined.

Believe I've substituted the correct link here.

@martinklepsch


martinklepsch:
Nick catch 👍

PR #180, opened 2015-08-06
Add echarts 2.2.7

huahaiy:
[ECharts] (http://ecomfe.github.io/echarts/index-en.html) is a Javascript visualization library. It is developed by Baidu (Chinese equivalent of Google) and open-sourced with a MIT license. The library is built on HTML Cavas, and it supports all major browsers. Unlike lower level libraries such as d3, ECharts allows declarative specification of charts using plain data, hence it is a very good fit for Clojurescript.

### DIFF
--- a/echarts/README.md
+++ b/echarts/README.md
@@ -1,0 +1,20 @@
+# cljsjs/echarts
+
+[](dependency)
+```clojure
+[cljsjs/echarts "2.2.7-0"] ;; latest release
+```
+[](/dependency)
+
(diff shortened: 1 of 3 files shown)


martinklepsch:
Sorry for taking so long Huaha, I just merged and published echarts to Clojars :)

huahaiy:
Thanks a lot!

chanshunli:
@huahaiy If i need use echarts 3 in ClojureScript, How can i do?

Issue #168, opened 2015-07-23
Question: Half finished externs


ducky427:
Hi,
I have created a partial extern file for [vis.js](https://github.com/almende/vis) visualisation library. I've written the externs for bits of the library which I am using.

Would you accept that for inclusion in this project?

Thanks

martinklepsch:
Hey @ducky427! Generally having externs as complete as possible is desired but not absolutely mandatory. Did you write the externs you already have by hand? Maybe this tool could help providing more complete externs? http://www.dotnetwise.com/code/externs/

ducky427:
@martinklepsch I tried using that site, but it gives me a alert saying "failed". So I had to write the extern by hand.

We are using a portion of that library and its been in production for over 6 months.

I've forked this repo and commited my work [here](https://github.com/ducky427/packages/commit/6ecfb31cacff4f183ffcaa8b3504230d86c6046d)

If its acceptable, I can create a PR.

Thanks

martinklepsch:
@ducky427 have you tried this one: https://github.com/jmmk/javascript-externs-generator as well?

Cheers

ducky427:
That worked! For some reason, the hosted version didn't work but I was able to run it locally and generate the complete extern file. Its [here](https://github.com/ducky427/packages/blob/master/vis/resources/cljsjs/vis/vis.ext.js).

Should I create a PR for this if its acceptable?

martinklepsch:
If you've tried using the jar in your project and things worked, sure go for it. :+1: 
Looking at your fork the nesting for the ext.js file is not completely right, needs to be in `resources/cljsjs/vis/common/` :smiley: 

Closing this and looking fwd to your PR :)

ducky427:
Thanks!

PR #182, opened 2015-08-09
[vis] Added vis v4.7.0

ducky427:
This is as discussed in #168


martinklepsch:
Took a while but finally merged and deployed! :)

ducky427:
Thanks a lot! :+1:

PR #186, opened 2015-08-13
Add virtual-dom

Szczyp:


crisptrutski:
Hey @Szczyp, thanks for the contribution

Note we already have #165 pending, but prefer that your version is tied to a github release rather than downloading via a branch.

Shortcomings of your PR are a missing README and also the externs are not quite as complete.

@martinklepsch tempted just to combine the best of both of the PRs and just merge now - sounds like putting `virtual-hyperscript/svg` in a separate package is the quickest way forward.

Szczyp:
Oh didn't see that. I use this internally and see vdom isn't in the repo so...

PR #191, opened 2015-08-17
Added marked markdown parser.

12foo:
Hi,

this is a pull request adding [marked](https://github.com/chjj/marked), a fast markdown parser with GFM. I tried using showdown first because it was in the list, but the cljsjs version of that seems to be very outdated for some reason.

Best,
Phil

### DIFF
--- a/marked/README.md
+++ b/marked/README.md
@@ -1,0 +1,17 @@
+# cljsjs/marked
+
+[](dependency)
+```clojure
+[cljsjs/marked "0.3.5"] ;; latest release
+```
+[](/dependency)
+
+Use in your app:
+
+```clojure
+(ns application.core
+  (:require cljsjs.marked))
+
+(def my-html (js/marked my-markdown))
+```
+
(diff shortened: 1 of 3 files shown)


crisptrutski:
Merged in bd70261d21ef86a9de5554973c8b017e29cf79a6 but published as 0.3.5-0 (see 178fd9ec5e6dd09f251f5a5fe7ceea906d94485c)

crisptrutski:
Thanks! 

Created #192 also so that interested parties may update the package.

The "some reason" is that there's nothing automatically tracking version.

@martinklepsch perhaps an optional task to detect the latest upstream version would be a good addition to the template? could perhaps have a small app that maintains badges for each library's readme based on that task

martinklepsch:
@crisptrutski
Might be handy but not sure if it's worth the work. Most work is in externs and for that we rely on contributors. 

I see CLJSJS as a platform where people can share stuff they need(ed) and so if the person who contributed Showdown needed it in that old version that's ok.

You've been helping lots of people with your contributions here so if you want and have the energy to push things in certain directions go for it. :beers:

crisptrutski:
@martinklepsch will see if some free time materialises :package: 
 
Anticipating 99% of packages can return a latest version number from NPM or Github, so would just need to add some helper tasks and add one-liners in the build.boot files.

1. Minimise surprise for uses who are not checking versions (assume that it's recent, ie. matches default documentation elsewhere)

2. Save time for users selecting between libraries

PR #197, opened 2015-08-21
Reactabular 0.6.6

mikesperber:
Depends on lodash, which I just submitted.


martinklepsch:
Is it possible that this package does not include the project's source files? 
A change like the one indicated in `build.boot` could be handled programatically as done in [the echarts package](https://github.com/cljsjs/packages/blob/master/echarts/build.boot#L27) and others.

mikesperber:
When I submitted the pull request, it needed a minor patch, which was why I didn't download the dist files. The patch was merged upstream yesterday, so I'll see if I can zap the files again and update the pull request.

mikesperber:
Done.

martinklepsch:
Thanks!

PR #202, opened 2015-08-26
Updated c3 checksum

kenfehling:


kenfehling:
Updated c3 checksum for zip file

martinklepsch:
Thanks!

PR #203, opened 2015-08-27
Nodejs externs 1.0.4

aJchemist:
I've tested nodejs externs through this. It worked in advanced compilation as I expected. thx for `boot-cljsjs` hard work!

### DIFF
--- a/nodejs-externs/README.md
+++ b/nodejs-externs/README.md
@@ -1,0 +1,15 @@
+# cljsjs/nodejs-externs
+
+[](dependency)
+```clojure
+[cljsjs/nodejs-externs "1.0.4"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` only. So there is no need to require and if it's
+simply listed as a dependency, the externs should be loaded at cljs compliation
+time.
+
+## References
+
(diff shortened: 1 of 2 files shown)


martinklepsch:
Thanks, this is a great addition. I'd suggest adding something to the readme that explains how to use the package, i.e. there is no need to require and if it's listed as a dependency the externs should be loaded.

Also wondering if the package name is ideal, maybe @crisptrutski has thoughts?

crisptrutski:
Name sounds good to me, `-externs` seems like a good naming convention for this kind of thing.

Interested in the motivations for advanced compilation for node.js and other server contexts. Are there noticeable performance wins?

martinklepsch:
AFAIK advanced brings many execution optimizations besides the plain filesize stuff. E.g. scripts compiled with `simple` will be slower than ones compiled with `advanced`.

aJchemist:
@martinklepsch ping. I've updated `README.md`.

martinklepsch:
Closed via https://github.com/cljsjs/packages/commit/285fe132ff31e6e473c39bd54ec0378b00cebb9f & https://github.com/cljsjs/packages/commit/26f00b558449b405ac1d4ee7ada61e19886493bc. @aJchemist I added a `-0` to the version so we can release patches for the same version :)

Thanks a lot! :+1:

PR #204, opened 2015-08-27
Add Material Design Lite

ClashTheBunny:
I'm using Material Design Lite in a project and figured I'd add it to cljsjs.


ClashTheBunny:
I just added a squashed commit with a note on the necessity to include material-icons.

ClashTheBunny:
Updated with Licensing info.  MDL is released under the Apache-2.0 license.  My code is MIT.

martinklepsch:
Hey Randall, thanks for your contribution. Some of the names seem to be out of sync. Sometimes the package is called `mdl` in other places it's `material`. I'm thinking maybe `material-design-lite` is the best name to avoid any confusions with potential other material design libs?

ClashTheBunny:
This is Google's `material` library, and the directions said to name it the same as the JS file.  The `mdl` is an artifact of renaming it after I added it all and found that `mdl.zip` extracts to `material.js`.

I've renamed all instances of `mdl` and of `material` to `material-design-lite` (even the JavaScript file from the package, is that right?).

What is the best way to test things like this?  I just have it in a temp project that I rebuild the temp project when I update material-design-lite.  It would be nice to have some guidance in the packaging howto.  Is it possible to just test in phantomjs or node if the externs are available when importing the library?  This way we are sure we are actually providing what we say we are providing.

martinklepsch:
> What is the best way to test things like this?

Using it in another project is what most people do for testing I believe. After you ran `boot package build-jar` you can add it to `:dependencies` in other projects.

Regarding the name I assumed that `-lite` is something different than the official Google release. If you feel `material` is most appropriate I'm happy with that.

ClashTheBunny:
I'd think, since it's the official Google implementation of material
design, that `material` alone or the full `material-design-lite` would be
the best.  They are the only material design library that are working with
the material design team and they get approval for all of their components
with them.  I think that renaming the .js file feels a bit hackey, but it's
no big deal for me.

martinklepsch:
Your call. You convinced me material is an equally valid choice :)

martinklepsch:
Thanks!

PR #206, opened 2015-08-29
Add paper.js to packages

ryankee:
Adding [Paper.js](http://paperjs.org) to the mix.


Deraen:
Looks good.

As with Wad, I don't see the reason to mention the license on the extern file. It's already mentioned on pom metadata. Also, acorn and straps are just some tools used to build Paper.js, that's not too interesting for us. Just trying to keep cruft to minimum, even on comments.

crisptrutski:
Same drill as #207, merged in 3c01eabea9771f222b895dbc71d0597ce424c566 as `[cljsjs/paperjs "0.9.24-0"]`

Thanks again for sharing, and for your patience :seedling::ear_of_rice::deciduous_tree:

PR #207, opened 2015-08-29
Add Wad to packages

ryankee:
Adding [Wad](http://www.codecur.io/us/songdemo).


Deraen:
Looks good.

About the license on the extern file. I'm not think it is required as the extern file is not from wad project?

crisptrutski:
Sorry this sat for a while, just noticed this two vetted PRs

Merged in ccee28be806631a5a454f61d3091a0747421c912, after some small amendments in d1ddab7760c0127cbe51eaf7387b6511d4cb977f, and published as `[cljsjs/wad "2.2.3-0"]`

Thanks for the contribution! :microphone::guitar::yum:

PR #210, opened 2015-08-31
[nodejs-externs] patch to 1.0.4-1

aJchemist:
@martinklepsch, oh I'm sorry. I've made a mistake.

originally I wrote regex like this (and I tested this `package build-jar`)
```
(sift :move {#"^node.js-closure-compiler-externs-1.0.4/([^/]*).js$"
                "cljsjs/nodejs-externs/common/$1.js"})
```
then secretly push force to this later ( untested :-( )
```
(sift :move {#"^node.js-closure-compiler-externs-.*/([^/]*).js$"
                "cljsjs/nodejs-externs/common/$1.js"})
```
BUT!! 

this moves unwanted contrib, tests, etc. externs (not core nodejs externs) to target.
and It will break down compilation. so I've fixed.

### DIFF
--- a/nodejs-externs/README.md
+++ b/nodejs-externs/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/nodejs-externs "1.0.4-0"] ;; latest release
+[cljsjs/nodejs-externs "1.0.4-1"] ;; latest release
 ```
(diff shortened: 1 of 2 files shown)


martinklepsch:
No worries, just deployed the fixed release :+1:

PR #215, opened 2015-09-03
[react-bootstrap] release 0.25.1

ducky427:
I've updated the dependency. I also re-generated the externs as documented in the extern file.


crisptrutski:
Nice job having the PR number be an anagram of the version :clap: 

Externs look a bit funny, with reordering and deletions - a niggle wonders if there hadn't actually been manual intervention with the file before.

Optimistically merged in f36e89b05b941fe1097e52feb319d5e1f7b7ee34 and released to clojars

ducky427:
Thanks a lot!

I had a cursorily investigated the difference. I noticed that `GLYPHS` has been removed from the latest `react-boostrap` package it seems. That produced a lot of removals.

Also, the latest package isn't visible on clojars yet. Is there a delay in becoming visible?

Cheers

crisptrutski:
Should be visible now :eyes:

ducky427:
cheers!

PR #216, opened 2015-09-03
Selectize update

mangolas:
Did my initial pull req bit too fast, only this actually works with advanced compilation mode. Sorry for extra work.


crisptrutski:
Very little "extra" work - we wouldn't be here if I'd tested first :wink: 

Pushed out as `0.12.1-1` in d146f0ba76c3680fd0edf67006594347725b2c5d

PR #218, opened 2015-09-05
Added moment-range.js - Range extenstion for momentjs

chadhq:
Hello,

I've added moment-range 

https://github.com/gf3/moment-range
http://gf3.github.io/moment-range/

It is a plugin extension for the already existing moment(js) library.

Thanks for considering! Let me know if I've forgotten to do anything...

### DIFF
--- a/moment-range/README.md
+++ b/moment-range/README.md
@@ -1,0 +1,30 @@
+# cljsjs/moment
+
+[](dependency)
+```clojure
+[cljsjs/moment-range "2.0.3"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
I presume moment-range depends on Moment. So `cljsjs/moment` should be listed as dependency of this package and `cljsjs.moment` added to `deps.cljs` `:requires` (something like `(deps-cljs :name "cljsjs.moment-range" :requires #{"cljsjs.moment"})` should do it).

Also, remove the commented out stuff from `build.boot`.

chadhq:
Great, thanks for the feedback, makes sense :-)

Assuming I've got this worked out now, I'm going to get started on the moment/timezone stuff as well.

Thanks again

Deraen:
Thanks. Merged, made small changes and released.

Deraen:
@chadhq Btw. while your are looking into Moment stuff, I think Moment.js extern could also use improvements: #6 :)

chadhq:
@Deraen Sure, I'll take a look.

Thanks for that.

PR #220, opened 2015-09-07
BankersBox 0.1.0

phronmophobic:
Adding BankersBox, https://github.com/twilio/BankersBox.

"A redis-like wrapper for javascript data storage using localStorage as the default persistent data-store."


crisptrutski:
Hi @phronmophobic, thanks for the submission!

To ease maintenance, we prefer to download the JS sources directly from a (versioned) URL rather than bundling them in this repo. Here's a good example: [d3/build.boot](https://github.com/cljsjs/packages/blob/master/d3/build.boot), and the BankersBox zip is [available here](https://github.com/twilio/BankersBox/archive/v0.1.0.zip)

Could you please update the PR? 

While you're at it, the externs look good at a glance, but from the README I get the impressions we should also include `BankersBoxLocalStorageAdapter` and `BankersBoxNullAdapter`.

phronmophobic:
Hey @crisptrutski,

Thanks for the quick response. 

After playing around with bankersbox more, it doesn't seem like BankersBox is even worth wrapping. It serializes all data for localStorage with JSON.parse and JSON.stringify which means that clojurescript values won't round trip. There is no builtin way to swap the serde part of bankersbox (the adapters don't help). 

Anyway, I was hoping that it would be less work to wrap BankersBox than to implement a similar API, but I'm not yet sure if it's worth mucking around with library's internals or writing/finding another implementation. 

Should I delete the pull request request for now unless it seems like BankersBox is worth wrapping?

crisptrutski:
I'm not sure what you're trying to achieve, but if you just want to store data, and aren't too attached to Redis data structures you may find [hodgepodge](https://github.com/funcool/hodgepodge) fits your needs, or one of its inspirations: [storage-atom](https://github.com/alandipert/storage-atom) or [plato](https://github.com/eneroth/plato)

crisptrutski:
As for what-to-do with this PR - there's not much work to finish it off, and little harm in adding it, but it's not something I'd see myself using either - seems a little bit on the naive side implementation wise

crisptrutski:
Perhaps looking into more mainstream client-side databases, and seeing if they have wrappers that suit you - like https://github.com/davber/clj-pouch

phronmophobic:
Alright, updated the pull request with the improvements you suggested. I'm still not sure it's worth merging in. Since I wouldn't recommend it for clojurescript use, I wouldn't want someone to consider its presence in cljsjs as a recommendation.

As for what I'm looking for, I would be happy with either something like Indexeddb or WebSql, but it looks like Web Sql is being deprecated, and IndexedDb isn't supported across the board yet. 

Pouchdb looks really interesting! I'll definitely check it out. Thanks for the pointer!

crisptrutski:
No harm in sharing this work - CLJS users aren't afraid of interop and coercing data :feelsgood: 

Released as `0.1.0-0`, no need to increment counter during that initial dev 

Thanks! :bank:

PR #221, opened 2015-09-10
add parse

jab:
Dear cljsjs maintainers,
I'm new to CLJ(S), and the first thing I wanted to try was building a CLJS app on top of [Parse](https://parse.com/). I found my way here and noticed e.g. Firebase is packaged but Parse is missing, so thought I'd try following [these docs](https://github.com/cljsjs/packages/wiki/Creating-Packages) and submitting a (heavily cargo-culted) PR. Not sure this is correct but hopefully you can help me fix it up if not. Thanks in advance for reviewing.

### DIFF
--- a/parse/README.md
+++ b/parse/README.md
@@ -1,0 +1,18 @@
+# cljsjs/parse
+
+[](dependency)
+```clojure
+[cljsjs/parse "1.6.0-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.parse))
+```
+
+[flibs]: https://github.com/clojure/clojurescript/wiki/Packaging-Foreign-Dependencies
--- a/parse/build.boot
+++ b/parse/build.boot
@@ -1,0 +1,28 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[adzerk/bootlaces   "0.1.9" :scope "test"]
+                  [cljsjs/boot-cljsjs "0.5.0" :scope "test"]])
+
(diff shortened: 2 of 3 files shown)


martinklepsch:
This looks pretty good already. We always add a `-0` to the version of initial releases which we increase when we release a new CLJSJS version of the same library version. So please add `-0` to your version. :)

Thanks for your contribution! :+1:

jab:
@martinklepsch Thanks for the quick review! Should have noticed about the `-0`, added now. (Edited the commit message too to match the format you use.) Also, noticed that Parse just released version 1.6.0 of the JS SDK so updated to that as well.

I've confirmed that I successfully end up with a `~/.m2/repository/cljsjs/parse/1.6.0-0/parse-1.6.0-0.jar` after running `boot build-jar`, but I have yet to successfully `require` it, probably because I'm doing something wrong. So probably worth doing that quick sanity check before merging. If you get a minute to look at this together on Slack, would be much appreciated, but no worries if not.

martinklepsch:
Have you tried the require call as it's in the README?
```clojure
(ns application.core
  (:require cljsjs.parse))
```
From what I can tell this should work. Maybe there is some edge case though. I'm on vacation from tomorrow so I suggest asking in the #clojurescript slack, maybe also @crisptrutski can help.

I'll be back on 17th :)

jab:
Thanks @martinklepsch, got it working! (Starting to understand dependencies in cljs and boot a bit better... to boot;)
I think this is safe to merge!

jab:
(but if someone with more experience than me wants to try this and make sure, definitely wouldn't hurt.)

crisptrutski:
@jab gave this a spin with a translation of the getting-started guide - and glad to report it *parsed* with flying colours. `cljsjs.parse "1.6.0-0"` is released, merged in 852622c6d6e152243a707b79b0ea7c3c7554a9ac

jab:
hehe, thanks for testing and merging, @crisptrutski!

PR #226, opened 2015-09-14
BitAuth with Unicode Support

xcthulhu:
I have also taken the liberty to place a browser bundle in the repo.

I plan to eventually incorporate ECDH shared secrets in their encryption system, which is just AES for now.

I will keep this up-to-date as it is an upstream dependency for my clj-bitauth effort.

### DIFF
--- a/bitauth/README.md
+++ b/bitauth/README.md
@@ -1,11 +1,18 @@
 # cljsjs/bitauth
 
 ClojureScript wrapper around BitPay's [BitAuth](https://github.com/bitpay/bitauth) protocol.
+
+To include in your `project.clj`:
+
 [](dependency)
 ```clojure
-[cljsjs/bitauth "0.2.1-0"] ;; latest release
+[cljsjs/bitauth "0.2.1-0"] ;; Official release
+[cljsjs/bitauth "0.2.2-xcthulhu"] ;; Fork with Unicode Support
 ```
 [](/dependency)
+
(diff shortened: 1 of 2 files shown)


xcthulhu:
My first commit was premature, I did not verify that my changes to BitAuth worked upstream.

I can corroborate that the version I have pushed here indeed works upstream :dancers:

crisptrutski:
Thanks for saving us publishers from compiling scary internet npm projects on our squishy machines :smile: 

Not sure how we want to handle forks of libraries here 
- Is a non-numeric build number really enough? 
- Isn't there still value in having a build number for forks also, ie. `0.2.3.xcthulhu-0`? 
- Maybe changing the package name is more explicit (`cljsjs/bitauth-xcthulhu`) - for now this is a fork, not an alpha

This last one is a bit awkward as usually the group (here `cljsjs` would be the portion changed).

Perhaps @martinklepsch will be back from being a Finnish celebrity soon to make such a call? :video_camera: :smile: 

Though if this situation is a special case, and you can confirm your changes landing, I don't mind cutting the red-tape and just shipping :shipit:

xcthulhu:
I could just just fork and rebrand this as something else, the BitPay people have been rather sluggish about maintaining their own open source product.

I'll probably call it nauthilus because I love terrible puns like that 🐙.

Anyway, I'm cool with whatever version naming scheme you guys run with.  I'll defer to @martinklepsch  to decide what to do I think.

crisptrutski:
@martinklepsch @xcthulhu just pinging on this before it gets too cold

xcthulhu:
Hey!

Me and the BitAuth peeps got together and we fixed everything upstream.

Let me make a new pull request reflecting this.

PR #237, opened 2015-09-22
Added NProgress (http://ricostacruz.com/nprogress/)

dhruvbhatia:


crisptrutski:
Thanks for the contribution!

One quick and valuable improvement would be to add a minified version for production builds ([an example](https://github.com/cljsjs/packages/blob/91f4f43cde5123c60b1878051964da2de3ce0dba/virtual-dom/build.boot#L26-L27))

Also it's probably worth including some boilerplate around [using css from cljsjs libs](https://github.com/cljsjs/packages/wiki/Non-JS-Assets) in the READMEs for visual projects (would you agree, @martinklepsch?)

The externs are structured in a way I haven't seen before.. had to test them to convince myself that they worked. Haven't seen this "functional" style of declaration before!

dhruvbhatia:
@crisptrutski Thanks for your feedback!

I've added a minified production version and improved the README to outline instructions for including the CSS file.

Re: the externs file, is there a preferred style this should be in? I copied [the Oboe package's style](https://github.com/cljsjs/packages/blob/master/oboe/resources/cljsjs/oboe/common/oboe.ext.js) and it works as intended, but happy to refactor if required.

martinklepsch:
I guess it would be cleaner to not define the externs toplevel but define them as properties of the exposed object. An example of this can be found [in the chance externs](https://github.com/cljsjs/packages/blob/master/chance/resources/cljsjs/common/chance.ext.js).

I'm not too worried about it though. 

Linking to the Wiki page on including non JS assets is a great idea and probably something that we should add to other packages using CSS. Thanks!

crisptrutski:
Thanks for the updates!

Just to echo Martin - have no qualms about this style of externs, was just commenting that you'd taught me something :smile:

dhruvbhatia:
Thanks guys!

Upon investigation, I found an [externs generator](http://www.dotnetwise.com/Code/Externs) so have updated as per that. :)

crisptrutski:
Thanks for your patience and flexibility, merged in 2ef34e486c41fead2bb94a397ca784fda45f72d8 :bar_chart: :beers:

Issue #229, opened 2015-09-15
Package more of closure-compiler contrib externs


Deraen:
Clousure-compiler repository contains some maintained extern files: https://github.com/google/closure-compiler/tree/master/contrib/externs

Currently google-maps-v3 and facebook externs are packaged and the extern is included in our repo. I'm not sure if it would be better in this case to download zip and extract the extern from that. Anyhow, we should have way to track which externs have been updated in closure-compiler repo.

The repository also contains some interesting externs which are not yet packaged:

- [ ] Chrome extensions (not sure if useful, is advanced optimization used for extensions?)
- [ ] Google analytics
- [ ] Google tag manager
- [ ] Google universal analytics
- [ ] Google youtube
- [ ] Youtube player
- [ ] Processing.js

- Lots of angular externs, which we might not be interested in
- Some externs for w3c proposed(?) JS APIs

jab:
you can check off google analytics as of #239 ;)

PR #239, opened 2015-09-22
[google-analytics] add package (2015.04.13-0)

jab:
Hey cljsjs folks, (still new to all this and) noticed there's no google-analytics here so thought I'd take a stab. Was happy to see Google Closure actually [provides externs](https://github.com/google/closure-compiler/tree/master/contrib/externs/google_universal_analytics_api.js) and Google Analytics actually provide [a changelog](https://developers.google.com/analytics/devguides/collection/analyticsjs/changelog?hl=en), so that's what I used.

But now I'm wondering:

- Google Analytics is so popular, I was surprised there's no cljsjs package for it already. Is there no point in having a cljsjs package of it for some reason I don't understand?
- It looks like there are some other libraries that https://github.com/google/closure-compiler/tree/master/contrib/externs provides externs for that aren't in cljsjs. Is it worth feeding these into cljsjs too, perhaps in some automated way?

Thanks in advance for any answers, and hope this PR is actually useful.

### DIFF
--- a/google-analytics/README.md
+++ b/google-analytics/README.md
@@ -1,0 +1,23 @@
+# cljsjs/google-analytics
+
+The latest release as of 2015-09-22 is "2015-04-13" according to the
+[the changelog](https://developers.google.com/analytics/devguides/collection/analyticsjs/changelog).
+
+This is converted to a semver for cljsjs like so:
+
+[](dependency)
+```clojure
+[cljsjs/google-analytics "2015.04.13-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.google-analytics))
+```
+
+[flibs]: https://github.com/clojure/clojurescript/wiki/Packaging-Foreign-Dependencies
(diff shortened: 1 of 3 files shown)


crisptrutski:
Point 1: They're all using Mixpanel :trollface:?

To hazard a guess, while GA is certainly popular, the major value ad is the "who" from impressions, and far less people are triggering events. Also it's hard to quantify how many codebases out there contain downloaded or self created externs, so don't read too much into an absence. There are also other conspicuous absences..

The philosophy of CLJSJS is less to curate "worthy" libraries than it is to provide "at most once" schlepp, possibly with a side orders of "see how nice boot is", and "see how easy open source is". 

Point 2: This has also been suggested here #229, it's a good idea!

Thanks for another useful contribution @jab

crisptrutski:
Merged in a8dcf1ed

jab:
Thanks for the quick review, response, and explanation!

Deraen:
I wonder if it's good idea to include the analytics.js in deps.cljs in this kind of library? Google might be updating the library constantly.

It's easy to only package the extern and leave it for user to include script tag for analytics.js.

crisptrutski:
Considering the convenience - people may also just want to just paste the official snippet into their markup rather than initialising in their own code, with the extra advantage of staying up-to-date as you mention. It doesn't seem to be updated that often, the latest is from April. Perhaps creating a externs package AND a js package (that requires) externs is worthwhile a few extra lines for the flexibility.

Regarding vendoring the js, perhaps that's a worthwhile. The checksum is already a defence, and doubt republishing old versions would ever come up, so I just let the unversioned link slide.

PR #246, opened 2015-09-28
[chosen] Add chosen - enhanced select box jQuery plugin

osbert:


crisptrutski:
Works like a treat, lift-off in 9b527d37a49af94155ce38daa4949db25aafb3b2 :rocket: 

Thanks for going the extra mile and giving pointers for using the resources :+1:

PR #248, opened 2015-09-30
[chartist] add package (0.9.4-0)

LauraSempere:


onetom:
The exports file is incomplete so far because http://www.dotnetwise.com/Code/Externs/ failed to generate them from https://cdnjs.cloudflare.com/ajax/libs/chartist/0.9.4/chartist.js

We haven't included the SCSS sources yet because we don't have experience with `boot-sass` or `boot-scss`.

crisptrutski:
Thank you Stewart Mackenzie, Tamas Herman and Laura Sempere :smile: 

Not fussed about asset sources, most cljsjs libs just include the built styles.

crisptrutski:
Merged in 664fafa9e17fb6fe70f7e3948c18ec40a14dd6c0

PR #250, opened 2015-10-08
Fix Codemirror mode tests overwriting modes in some cases

djwhitt:
The mode js regex was matching both modes and their tests. When tests
were present this was causing the mode test files to overwrite the modes
themselves sometimes (depending on sort order I think).


crisptrutski:
Perfect! :dart:

djwhitt:
Wow, fast response. Thanks!

PR #255, opened 2015-10-10
[bitauth] Bumping to version 0.3.1

xcthulhu:
BitAuth now had unicode support!

With this revision I will be updating my wrapper accordingly.

### DIFF
--- a/moment/build.boot
+++ b/moment/build.boot
@@ -5,7 +5,7 @@
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
 (def +lib-version+ "2.10.6")
-(def +version+ (str +lib-version+ "-2"))
+(def +version+ (str +lib-version+ "-3"))
 
 (task-options!
   push {:ensure-clean false}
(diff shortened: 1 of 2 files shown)


xcthulhu:
Okay, I have been working on this, it appears that BitAuth 0.3.1 broke Closure support by switching how it was using *browserify*.

This was breaking CLJSJS and my own ClojureScript wrapper further downstream.  I believe that my latest commit bitpay/bitauth#56 restores CLJSJS and my wrapper, but it's not 100% clear.  The javascript code being wrapped is output by browserify which is complicated and unpredictable.

I will wait for BitPay to pull my commit in, and make another pull request when I can confirm using ClojureScript testing that the latest version of bitauth is functions properly with clojurescript.

xcthulhu:
I can confirm that this version has been tested with my UTF-8 branch of my clj-bitauth project and this is 100% compatible with my JVM implementation.

Deraen:
I won't accept PR which adds the source code to the repo.

PR #265, opened 2015-10-15
[react] React 0.14.0 extern fixes

favila:
* Change react-dom extern path to match packaging guidelines.
* Adds missing extern symbols (especially in react-dom).
* Removes some symbols removed by React 0.14.0 (not done thoroughly).
* Marks @depreciated some symbols depreciated by React 0.14.0.


r0man:
Hello, is there anything left to do on this PR? I'm interested in these extern definitions and rely on them in some of my tests.

martinklepsch:
Looks good, will merge now.

martinklepsch:
should be released now! thanks @favila! :+1:

r0man:
Cool, thanks!

r0man:
@favila and @martinklepsch I'm trying to access js/ReactDOMServer but get this in my Node REPL:
```
ReferenceError: ReactDOMServer is not defined
    at repl:1:89
    at repl:9:3
    at repl:14:4
    at Object.exports.runInThisContext (vm.js:54:17)
    at Domain.<anonymous> ([stdin]:41:34)
    at Domain.run (domain.js:221:14)
    at Socket.<anonymous> ([stdin]:40:25)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
```
I'm not sure how to require that JS. Is the server rendering JS even included in this package?

Roman.

martinklepsch:
@r0man do you have a depdency on `cljsjs/react-dom`?

r0man:
@martinklepsch Yes, I have that. ``[cljsjs/react-dom "0.14.0-1" :scope "provided"]``.
I also require ``cljsjs.react`` and ``cljsjs.react.dom`` at the top of my namespace. But I have the impression this doesn't pull in the JS that defines ``ReactDOMServer``.

r0man:
I can't find a reference to ReactDOMServer in the minified JS. I guess it's in another package. Which would make sense, since this is for server rendering.
https://facebook.github.io/react/docs/top-level-api.html#reactdomserver

martinklepsch:
@r0man Strange indeed. The sources don't include it but also there's no other separate package. Maybe it helps to look into the files coming with this one: https://www.npmjs.com/package/react-dom

r0man:
Thanks Martin, I'll take a look tomorrow ...

favila:
@r0man ReactDOMServer methods are actually defined in react.js, but exported via magic property trickery by the react-dom.js. (Same with ReactDOM.) However the prebuilt react-dom.js does not expose ReactDOMServer, probably because it is built for browsers only.

react-dom.js could be patched manually to expose the ReactDOMServer object. Long-term perhaps we need to start building react ourselves instead of downloading the prebuilt from the cdn?

joshfrench:
Is there a workaround for making `ReactDOMServer` available?

PR #273, opened 2015-10-28
Add packages react-input-autosize and react-select

wavejumper:
This pull request adds packages for [react-select](https://github.com/JedWatson/react-select) and [react-input-autosize](https://github.com/JedWatson/react-input-autosize) (a dependency of react-select)

I had to fork react-select because of https://github.com/JedWatson/react-select/issues/526  - so the package currently points to my fork.


olivergeorge:
Hi, just looking to try react-select.  I'll give this a go.  

I note that the ticket you mentioned is resolved... that might mean we don't need a forked repo for this.

wavejumper:
Correct. When a release is pushed with this change in place I will update my pull request :)

Deraen:
What's the status? Does this still refer to a fork?

atroche:
Yep, and it looks like the fix for https://github.com/JedWatson/react-select/issues/526 won't be available until a new release of react-select. Unless we're willing to base the cljsjs package off master…

Deraen:
I have merged #310 which added the same packages with more recent versions. Hopefully this can be closed?

atroche:
good work =) makes sense to me

PR #278, opened 2015-11-01
[responsive-nav] Added Responsive Nav (http://responsive-nav.com)

dhruvbhatia:


pandeiro:
I think we need externs for this, right? I wasn't able to get the externs generator to work for this lib for some reason, but the [public API is documented here](https://github.com/viljamis/responsive-nav.js#public-methods) and would be a good starting point.

dhruvbhatia:
Hi @pandeiro - I've tried using this package within a project with `:optimizations :advanced` on, and it compiles/works as expected. The `responsiveNav` function gets populated with all the correct public API methods, even though I haven't explicitly included them in the externs, [as shown in the output here](https://gist.github.com/dhruvbhatia/52a495350e56088ba09a).

However, let me know if you'd like me to explicitly specify them in the externs file and I can update if preferred.

Deraen:
@dhruvbhatia What is the source for the example? I'm not sure if that example shows anything.

dhruvbhatia:
@Deraen Apologies for the lack of clarity.

I've now created a [sample project](https://github.com/dhruvbhatia/responsive-nav-cljsjs-sample-project) which illustrates that the package works under advanced compilation and has all the correct API methods [(hence we don't need to explicitly specify them in the externs file)](https://github.com/dhruvbhatia/packages/blob/b8f9a29b63d8f0633b2a9768429702d028900ee6/responsive-nav/resources/cljsjs/responsive-nav/common/responsive-nav.ext.js).

You can simply clone the sample repo and follow the README instructions to see a working example: https://github.com/dhruvbhatia/responsive-nav-cljsjs-sample-project

It seems that the Google Closure Compiler doesn't compress the plugin's API methods (even under advanced mode), so they are all available for us to use. Hope this makes it clearer - let me know if you have any other questions.

Deraen:
@dhruvbhatia Closure never touches the source of external library. The extern files purpose is to make it possible to call external methods from Closure/ClojureScript code.

Try calling `(.toggle (js/responsiveNav))` and other methods.

dhruvbhatia:
@Deraen Hmm, that work fine for me  - though the syntax would be `(.toggle (js/responsiveNav "css_classname_of_nav_menu"))` since the responsive nav object first needs to be instantiated.

I've updated my example showing the `toggle()` function being called every 2 seconds - it works as I'd expect: https://github.com/dhruvbhatia/responsive-nav-cljsjs-sample-project/blob/master/src/responsive_nav_cljsjs_sample_project/cljs/core.cljs

Deraen:
That's possibly because prototype externs are global and the `toggle` method is defined by some other extern file. Still, I think the extern file for this lib should define all the methods used by lib, even if the methods already work.

Try calling `.destroy`, it should not be defined in the default externs included in Closure Compiler.

dhruvbhatia:
@Deraen Thank you, that's exactly what it was! I've now updated the externs file as per the plugin's public methods.

Deraen:
I wonder if the methods should be defined under `responsiveNav.prototype`. But if it works, no need to change, I see there are few other externs where methods are defined to the object directly instead of the prototype.

dhruvbhatia:
Thanks again @Deraen.

I managed to get the [externs tool working](http://jmmk.github.io/javascript-externs-generator/) by using [this source file](https://gist.githack.com/dhruvbhatia/4ab3e7c5535453c5fcba/raw/ba17a3847cecd7ba273789db88dc39788cbe87d1/responsive-nav.js) and have updated the externs file accordingly (it now uses `.prototype` and includes a few additional private methods that weren't in the plugin's documentation). I have tested all methods under advanced compliation and they all work as expected.

Thanks again for your help on this. :+1:

Deraen:
Deployed

PR #280, opened 2015-11-02
Fix Typo In Codemirror README

zindlerb:
Seems like it should be now instead of know.


martinklepsch:
fixed in master, thank you @zindlerb

PR #285, opened 2015-11-06
Added first shot at bootstrap-dialog.

ulsa:
It's not downloading and unzipping, which would be cool, and I'm not 100% sure about the externs file, but it's something.


Deraen:
Please download the source using Boot instead of including the sources in the repo.

Deraen:
Closing as inactive.

PR #286, opened 2015-11-06
fix .getDoc and move 2 methods to Doc Object

continuata:
- most of the CM methods in this extern are incorrect, as they assume
they are methods on the CodeMirrorObj, when they are actually methods
on the CM Document Object
- added CodeMirrorDoc definition and redefined 2 of the methods
(getRange, replaceRange)
- Other methods still need doing properly

See CodeMirror API for further methods affected http://codemirror.net/doc/manual.html#api

### DIFF
--- a/codemirror/README.md
+++ b/codemirror/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/codemirror "5.7.0-1"] ;; latest release
+[cljsjs/codemirror "5.7.0-2"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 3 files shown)


martinklepsch:
Thank you very much Stu! Merged and deployed to clojars :+1: 
```clojure
[cljsjs/codemirror "5.7.0-2"]
```

martinklepsch:
The release I pushed is broken :weary: Will push another release in a sec.

martinklepsch:
```
[cljsjs/codemirror "5.7.0-3"]
```
should be on clojars now.

PR #287, opened 2015-11-06
rename all occurences of React.ReactComponent to React.Component

thheller:
In React the ES6 class is available under `React.Component` and not `React.ReactComponent`.

Not 100% sure but I think `React.ReactComponent` is incorrect. I ran into this when `:advanced:` compiling something and some `React` functions got renamed on my components.

`om.next` ships with a basically empty externs that hides this issue [1], but I think if the name is changed to problem should go away.

Would be nice if someone could verify.

[1] https://github.com/omcljs/om/blob/master/src/main/om/externs.js

### DIFF
--- a/react/resources/cljsjs/react/common/react.ext.js
+++ b/react/resources/cljsjs/react/common/react.ext.js
@@ -33,19 +33,19 @@ React.isValidClass = function(componentClass) {};
 React.isValidElement = function(object) {};
 
 /**
- * @param {React.ReactComponent} container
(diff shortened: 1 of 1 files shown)


thos37:
This fix is critical for using cljsjs.react-0.14.0 in production

martinklepsch:
@thos37 I haven't merged this yet because I was waiting for confirmation that this change is correct, understanding your comment here as confirmation I merged and pushed it now.

PR #289, opened 2015-11-10
upgraded to react-bootstrap-0.27.3 and cljsjs.react-14.0

thos37:
tested fine with a fairly minimal app (Button, Panel, collapsible NavBar all work as before) and the new NavBrand works now.


thos37:
will submit a new PR that more closely matches the contributing guidelines

PR #293, opened 2015-11-11
[react-bootstrap] upgrade to 0.27.3 

thos37:
depends on [cljsjs/react-with-addons "0.14.0-1"] and [org.webjars.bower/bootstrap "3.3.5"]


thos37:
eh, still could be better ...

PR #264, opened 2015-10-15
[react-bootstrap] Update to v0.27.1

esessoms:
The old 0.25.2 version is no longer working with the new react 0.14.0 releases.  See https://github.com/react-bootstrap/react-bootstrap#react-v013-support

Most of the changes in the externs file seem due to ongoing library reorganization.

PR #294, opened 2015-11-11
[react-bootstrap] update deps to cljsjs.react 0.14.0-1 and fix a comment

thos37:
This PR builds on a previous open PR, https://github.com/cljsjs/packages/pull/264, which is needed to be compatible with the version of React (0.14) that's currently in CLJSJS.


Deraen:
Deployed.

PR #297, opened 2015-11-14
[react] fixes :optimizations :advanced errors 

thos37:
This PR is just a version number bump to enable merging PR #287 which is critical for production as it fixes serious errors that arise with :optimizations :advanced.  For example in Om.Next, when referring to React.Component, it's been optimized away:
Uncaught #error {:message "No queries exist for component path (#object[xA \"function xA(){React.Component.apply(this,arguments);this.state=null!=this.bb?this.bb():{};return this}\"])", :data {:type :om.next/no-queries}}

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -1,7 +1,7 @@
 # cljsjs/react
 
 ```clojure
-[cljsjs/react "0.14.0-1"] ;; latest release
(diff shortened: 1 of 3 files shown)


martinklepsch:
Will merge #287 and cut a release now.

PR #298, opened 2015-11-14
[vis] Added missing externs

ducky427:
I've added missing externs using [javascript-externs-generator](https://github.com/jmmk/javascript-externs-generator) which generated more complete externs.


ducky427:
thanks a lot! :+1:

PR #306, opened 2015-11-25
[react-day-picker] Add package

atroche:


Deraen:
Please define version in build.boot using `+lib-version+` and `+version+` vars:
```clj
+(def +lib-version+ "1.1.5")
+(def +version+ (str +lib-version+ "0"))
```

PR #308, opened 2015-11-27
Add missing externs for innerTickSize() and outerTickSize().

jszakmeister:
Bump the package release number to make this ready for release.

### DIFF
--- a/d3/README.md
+++ b/d3/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/d3 "3.5.7-0"] ;; latest release
+[cljsjs/d3 "3.5.7-1"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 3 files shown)


jszakmeister:
FWIW, I did run build this locally, and it resolved my issue with `innerTickSize()` and `outerTickSize()` being renamed.

It looks like the CircleCI build is broken for other reasons.  My best guess is that the gpg.conf needs to be updated to include `batch\n` in it.  If you scroll down in the [GPG Esoteric Options](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Esoteric-Options.html) section to `--passphrase`, you can see it has a note saying:

> Note that this passphrase is only used if the option --batch has also been given. This is different from GnuPG version 1.x.

jszakmeister:
I tried adding batch, but that didn't help.

Deraen:
CI build is broken because it's waiting for new Boot release.

Deraen:
Deployed

jszakmeister:
> CI build is broken because it's waiting for new Boot release.

I managed to see the issue about an hour ago.  Thanks for the info!

> Deployed

Thanks for the quick turn-around!

PR #314, opened 2015-12-01
Package react-quill 0.3.0

mbertheau:
I couldn't really confirm whether this works but it looks a lot like reactabular, so I suspect the error on my side somewhere else.

I'd appreciate it if someone could have a look. Thanks!

### DIFF
--- a/react-quill/README.md
+++ b/react-quill/README.md
@@ -1,0 +1,18 @@
+# cljsjs/react-quill
+
+[](dependency)
+```clojure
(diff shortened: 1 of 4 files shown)


Deraen:
Looks okay but you should really test this locally before this is merged.

mbertheau:
After adding the quill dependency I have it basically working now. No other changes were necessary.

mbertheau:
Thanks for merging!

Deraen:
I added build identifier to the version so this is deployed as `0.3.0-0`.

PR #322, opened 2015-12-05
Initial wrapping of three-water-material

timothypratley:
Hi,
This is a pretty niche library, but I found it useful... it provides a nice water material for threejs scenes.


Deraen:
Unless there is some reason, it think this should use the same name as original packaging uses: `ocean`.

timothypratley:
@Deraen would three-ocean be O.K.?
I'm fine with ocean, just suggesting it will be easier for people to find threejs related wrappers, seeing in this case it is highly threejs specific... what do you think?

Deraen:
@timothypratley I guess it's okay. I still would prefer just `ocean`. Description already mentions Three.js.

timothypratley:
O.K. just `ocean` it is :) will rename.

PR #327, opened 2015-12-09
[pikaday] add minified versions of js and css

pandeiro:
- minifies js and css and provides development and production output
- adds lib segment to all output paths
- bumps lib package version

I've tested this locally and it works but I'd like a quick review from somebody before publishing.


pandeiro:
This is working in both dev and production with the minified versions. Merging and deploying.

PR #330, opened 2015-12-10
First pass of paho lib.

thomas-shares:
Hi,

This is my first attempt at creating a pull request to CLJSJS, please review and comment and let me know what I can improve. 

Many thanks,
Thomas


Deraen:
I just updated the wiki with notes about version vars: https://github.com/cljsjs/packages/wiki/Creating-Packages#51-buildboot

I think the methods and properties are declared are incorrectly in the extern. I think the correct way would be to declare them as prototype methods of client: `Paho.MQTT.Client.prototype.connect = function(connectionOptions) {};`

Some examples from React extern:
https://github.com/cljsjs/packages/blob/master/react/resources/cljsjs/react/common/react.ext.js#L111
https://github.com/cljsjs/packages/blob/master/react/resources/cljsjs/react/common/react.ext.js#L150

Deraen:
@thomas-shares Hi, what's the status of this PR?

thomas-shares:
Hi,

Could this pull request be merged now? Thank you.

Thomas

Deraen:
Tests were still failing but I guess that's because this was based on such old commit. I'll see that this will get deployed.

PR #332, opened 2015-12-15
[pouchdb] 5.1.0

fehrenbach:
This just bumps the version.

I haven't been following cljsjs for a while. Have things changed significantly?

### DIFF
--- a/pouchdb/README.md
+++ b/pouchdb/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
No significant changes in Cljsjs.

Based on changelog, there has been some breaking API changes in PouchDB 3 -> 4 -> 5 so the extern probably requires some updates.

fehrenbach:
Good point. I went through the changelog and adapted the externs file. Fortunately, most changes seem to be on the data side, not the shape of the API.

PR #340, opened 2015-12-25
Add react-dom as dependency for react-select

schmee:
What is says on the tin. Without this dependency you get `Uncaught TypeError: Cannot read property 'findDOMNode' of undefined` when navigating the select box with the arrow keys, which leads to broken scrolling.

I haven't tested this locally, since `boot package build-jar` fails with

```
➜  react-select git:(react-select-dom) boot package build-jar
             clojure.lang.ExceptionInfo: java.lang.IllegalArgumentException: No such task (build-jar)
    data: {:file
           "/var/folders/t2/bmq0l4f516789351wk6b_sgh0000gn/T/boot.user3216260690496726718.clj",
           :line 21}
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: No such task (build-jar)
     java.lang.IllegalArgumentException: No such task (build-jar)
          boot.core/construct-tasks  core.clj:  754
                                ...
                 clojure.core/apply  core.clj:  630
                  boot.core/boot/fn  core.clj:  802
clojure.core/binding-conveyor-fn/fn  core.clj: 1916
                                ...
```

Any clue as to why this doesn't work?

### DIFF
--- a/react-select/build.boot
+++ b/react-select/build.boot
@@ -1,7 +1,7 @@
 (set-env!
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.0"  :scope "test"]
-		  [cljsjs/react "0.14.3-0"]
+                  [cljsjs/react "0.14.3-0"]
                   [cljsjs/classnames "2.1.3-0"]
                   [cljsjs/react-input-autosize "0.6.5-0"]])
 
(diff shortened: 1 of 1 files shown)


schmee:
Just realized that I should probably bump `+version+` as well?

Deraen:
Yes.

There is no longer `build-jar` no longer as bootlaces is no longer used, you can use `boot package pom jar install` to install the package locally.

schmee:
Done!

Ahh, I see. Just tested it locally and everything seems to be working. Mind if I update the wiki with this info?

Deraen:
I did already update one mention in wiki but feel free to check and update it.

schmee:
Great, thanks!

PR #347, opened 2016-01-02
cljsjs package for opentype

zachcp:
cljsjs package for [opentype js](http://opentype.js.org/)


Deraen:
Deployment failed because the checksum was wrong. I have now fixed it but please test packaging in future.

zachcp:
sorry @Deraen; thanks

PR #348, opened 2016-01-02
cljsjs for labella.js

zachcp:
cljsjs package for [labella](http://twitter.github.io/labella.js/)

### DIFF
--- a/labella/README.md
+++ b/labella/README.md
@@ -1,0 +1,18 @@
+# cljsjs/labella
+
+[](dependency)
+```clojure
+[cljsjs/labella "0.1.1-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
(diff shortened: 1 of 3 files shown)


Deraen:
If labella requires d3 there should be a maven dependency to cljsjs d3 package in addition to deps.cljs require.

zachcp:
added the d3 dependencies. Quick question: don't the cljs SIFT commands include the D3 files in D3's build.boot? isn't it redundant to also do the maven step?

Deraen:
Hmm what do you mean?

When building this package, fileset contains only the files from resources-paths and files added by the tasks (files extracted from the zip) so no d3 files will get included. in labella jar.

When one uses labella, the deps.cljs will refer to cljsjs.d3 namespace so that needs to be available in the project using labella. Maven dependencies are used to include cljsjs d3 package which includes another deps.cljs file which tells which files provide the cljsjs.d3 namespace and externs.

PR #350, opened 2016-01-04
Adds Autobahn|JS 0.9.9  as package

Ninerian:
I created a package for Autobahn|JS, a WAMP Client Library. Locally tested. Installed it with `boot package pom jar install`


aft-luke:
Has anyone been able to get this to work? I always get "no such namespace cljsjs.autobahnjs" errors.

PR #356, opened 2016-01-06
[jquery-daterange-picker] Update version

ayazmaroof:
Version 0.0.9 includes bug fixes and enhancements
Also updated moment and jQuery version

Some of the bug fixes:
- Auto close
- Range highlighter

https://github.com/longbill/jquery-date-range-picker/releases

### DIFF
--- a/jquery-daterange-picker/build.boot
+++ b/jquery-daterange-picker/build.boot
@@ -1,12 +1,12 @@
 (set-env!
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.0" :scope "test"]
-                  [cljsjs/jquery    "1.8.2-2"]
-                  [cljsjs/moment    "2.6.0-3"]])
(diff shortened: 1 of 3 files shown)


Deraen:
Hmm, interesting that this package includes the sources instead of downloading them. Anyway, I'll merge this and do another release soon which refactors the packaging task for this lib.

Deraen:
@ayazmaroof Hmm, upstream doesn't have 0.0.9 release so what version is this?

Deraen:
In addition to this `0.0.9-0` there is also now version `0.0.8-0` deployed which downloads the files from upstream. It also removes minified JS code as one is not provided upstream and the code can't be easily minified with asset-minifier.

ayazmaroof:
@Deraen I noticed the 0.0.8 release after I submitted the pull request. The 0.0.9 source code was available in the master of upstream. 
Thanks for updating and downloading the files from upstream.

PR #357, opened 2016-01-07
[blockapps] adds blockapps 2.2.5

pelle:
https://github.com/blockapps/blockapps-js


Deraen:
Looks like the build is currently broken because browserify can't be executed.

Deraen:
@pelle Is there a reason why browserify npm script (https://github.com/blockapps/blockapps-js/blob/master/package.json#L67) is not used?

Deraen:
I updated the build to use npm script.

PR #361, opened 2016-01-12
[incremental-dom] - Add incremental dom

tommichaelis:
This adds [incremental dom](https://github.com/google/incremental-dom) to cljsjs.

Incremental dom is google closure compiler compatible, so this package plays nicely with advanced compilation. However, the default incremental dom build uses [goog.module](https://google.github.io/closure-library/api/namespace_goog.html#module) to declare the namespace, which doesn't seem to be compatible with the clojurescript compiler's dependency resolution. I've included a patch for the incremental dom gulp file which outputs code which is compatible with the clojurescript compiler.

I'm not sure what the build server supports, but this will require nodejs and npm to build the library. If that's not supported, I'd welcome any suggestions as to the best way to get this into cljsjs.


Deraen:
Hmm, interesting. CI should support Node and npm.

I would like to check if we can get ClojureScript to support `goog.module` definitions, with a quick glance doesn't look like too hard. Could you provide me a version without the patch for me to test, for example in another PR?

I think without patch there is no need to use Make.

Deraen:
Adding Cljs support for `goog.module` is tracked here: http://dev.clojure.org/jira/browse/CLJS-1543

tommichaelis:
I've added a pull request which uses `goog.module` (https://github.com/cljsjs/packages/pull/362).

As an alternative, the original library is written using ES2015 modules, and also has gulp tasks to compile to CommonJS and AMD modules. I don't think any of those module systems are supported in the clojurescript compiler either, but if there's more impetus behind adding any of these it might be worth trying to integrate one of those options.

tommichaelis:
I've just updated this to use just boot. It still uses the patched gulpfile, but I did enough reading around the boot api to get it to work without shelling out to `make`.

Deraen:
Great! As adding support for `goog.module` in Cljs compiler wil take some time, lets merge this.

Could you just fix the header in readme :)

tommichaelis:
Fixed - cheers :)

Deraen:
I did just once change, I added build identifier to version: `-0`.

Should be deployed now, there was some errors from gulp related to unit-phantom, but I don't know if that matters? Probably CI has older PhantomJS version or something like that.

tommichaelis:
Sounds good, just tested it and it all seems to work as expected!

Cheers Deraen

Deraen:
Thanks for contributing the package!

PR #362, opened 2016-01-13
Adding incremental dom with goog.module declaration

tommichaelis:
Adds incremental dom, while leaving the goog.module declaration.

This uses the gulp task in incremental dom, rather than adding a new one (see https://github.com/cljsjs/packages/pull/361), so the resultant javascript file uses `goog.module` rather than `goog.require`. This is not currently supported by the clojurescript compiler.


Deraen:
Now that merged version is cleaned, it will be easy enough to remove the patch from it.

PR #363, opened 2016-01-13
Added a package for react-timer-mixin

ducky427:


react-timer-mixin/README.md
```diff
@@ -0,0 +1,18 @@
+# cljsjs/blockapps-js
```
Deraen: This line refers another package.

ducky427: Opps. I'll fix this.

ducky427: I've fixed that and squashed that commit into 1.

ducky427:
Thanks @Deraen!

PR #370, opened 2016-01-18
[react-timer-mixin] Use uglifyjs instead of boot task for minifying code

ducky427:


martinklepsch:
Mind outlining whats the benefit of using uglify?

ducky427:
@martinklepsch, I had to use uglify as the boot task uses the default config of language_in for asset-minifier which is es3. there was no handle in that boot task to modify it.

ducky427:
Using the `minify` task, I get this error:

```
SEVERE: ..../react-timer-mixin/16pc/f0sqpx/cljsjs/react-timer-mixin/development/TimerMixin.inc.js:78: ERROR - Parse error. IE8 (and below) will parse trailing commas in array and object literals incorrectly. If you are targeting newer versions of JS, set the appropriate language_in option.
  cancelAnimationFrame: _cancelAnimationFrame
```

martinklepsch:
@ducky427 Maybe this could be improved in the `minify` task? It seems the `minify-js` function in asset-minifier has a language option. https://github.com/yogthos/asset-minifier/blob/master/src/asset_minifier/core.clj#L104

In any case I'm ok with this change but maybe in general making the task a bit more customizable might make sense.

ducky427:
Sure. I am looking at the code for [`minify`](https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L134) and it can be done. But I wasn't sure about it and it may have been out of scope for this PR.

nha:
Other than that, I am working on a `uglify` boot task. The uglifying part works alright, but I still have to make it play well in `boot` (ie. really replace files) : https://github.com/nha/boot-uglify

PR #377, opened 2016-01-20
Update long.js and bytebuffer.js.

chadharrington:
- Update long.js to latest upstream lib. 
- Update bytebuffer.js to use latest long.js.


chadharrington:
Thanks for the quick turnaround!

Deraen:
No problem.

Had some hicupps with the automated CI as it tried build bytebuffer first and failed because long wasn't available yet :D

PR #378, opened 2016-01-20
Upgrade react-datepicker

AshtonKem:
I'm getting errors on the old version, checking to see if a newer version fixes it.

### DIFF
--- a/react-datepicker/build.boot
+++ b/react-datepicker/build.boot
@@ -7,7 +7,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "0.15.2")
+(def +lib-version+ "0.18.0")
 (def +version+ (str +lib-version+ "-0"))
 
 (task-options!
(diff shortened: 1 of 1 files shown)


Deraen:
Thanks for the contribution, but I had to revert this. The package task is failing and I can't easily fix it because the zip file does not contain plain JS files anymore, only JSX files. Probably it would require running some NPM commands to run JSX/Browserify to build the plain file. Check e.g. forge package task for how to run Node during package: https://github.com/cljsjs/packages/blob/e7abebb298acc839ec0145c9ebf93a092bf2c6c6/forge/build.boot#L32

Please test changes locally in future before opening a PR.

ducky427:
@AshtonKem, have a look at this issue Hacker0x01/react-datepicker#303. Seems like the repo doesn't have a dist folder anymore.

ducky427:
@AshtonKem, generating `dist/react-datepicker.js` and `dist/react-datepicker.min.js` is relatively straightforward:

- `npm install`
- `grunt webpack`

The first command may complain if ruby and `scss-lint` aren't installed.

But generating the css file is more involved as it involves having ruby and [`scss-lint`](https://github.com/brigade/scss-lint) installed.

AshtonKem:
I was under the impression that cljsjs packages don't expose css files for its users. If that's the case, does it matter if the case files aren't generated?

--
Ashton

ducky427:
Currently `react-datepicker` does expose the css files for its users which can be used by `boot` but not using `lein`.

@Deraen whats the policy regarding css files?

Deraen:
If the package has css files, it's good to expose them. Other packages should contain examples about this.

CI system has Ryby available so it should be possible to run a gem command before running grunt build, to install `scss-lint`.

There is no reason why Lein users could not make use of those css files, they just need either smart enough plugin (less4clj) or some middleware to serve them.

PR #386, opened 2016-01-28
Add FileSaver.js

theasp:


Deraen:
As far as I can see, https://github.com/eligrey/FileSaver.js is upstream version. Only difference with Teleborder version seems to be bower.json.

Deraen:
Upstream version doesn't have any releases, but you can download zip for speicifc sha: https://github.com/eligrey/FileSaver.js/archive/230de7d1426d0d04abe782e0aed5800838892dd7.zip

FileSaver.js mentions `1.1.20151003` as version.

Deraen:
And I think proper sha to use would be one where version number was updated: https://github.com/eligrey/FileSaver.js/commit/f23c4a66dc9fa1af818322f08eb29a371ccb4ccf (to get zip url, click browse files and download).

theasp:
You are correct, I chose to use the fork due to it having actual releases.  I have updated it to match your comments above.

theasp:
*sigh*

Ok, it's good now. Let me know if you want anything further done to it.

Thanks

Deraen:
Should be good now.

To check if packaging works you can run `boot package` locally :)

PR #388, opened 2016-01-30
[react-motion] Update to 0.4.1

edannenberg:
This PR still has a few issues:

* The react-motion release archive no longer includes the built js file, apparently it is now only served via `npmcdn.com`. I changed the package task to download from there but the task fails with:

    clojure.lang.ExceptionInfo: javax.net.ssl.SSLException: Received fatal alert: internal_error

Not sure if this an error on my end or with java and `npmcdn.com`. I did however put the js file on my own http server for testing and the download task worked as expected.

* The minify task fails on my rig with:

    clojure.lang.ExceptionInfo: java.lang.IllegalArgumentException: No implementation of method: :file of protocol: #'boot.tmpdir/ITmpFile found for class: nil

Looks like it can't find the input file, the sift task correctly moved the file though.

Removing the minify task allowed me to build and test the jar with a simple reagent project. Everything seems to be in order.

On a side note I had to include `clj-http` dependency, missing dependency of the `boot-cljsjs` task?

### DIFF
--- a/react-motion/build.boot
+++ b/react-motion/build.boot
@@ -1,11 +1,15 @@
 (set-env!
   :resource-paths #{"resources"}
-  :dependencies '[[cljsjs/boot-cljsjs "0.5.0"  :scope "test"]
-                  [cljsjs/react       "0.13.3-1"]])
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.1"  :scope "test"]
+                  [cljsjs/react       "0.14.3-0"]])
 
-(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+(require '[cljsjs.boot-cljsjs.packaging :refer :all]
+         '[boot.core :as boot]
+         '[boot.tmpdir :as tmpd]
+         '[clojure.java.io :as io]
+         '[boot.util :refer [sh]])
 
-(def +lib-version+ "0.3.0")
+(def +lib-version+ "0.4.1")
 (def +version+ (str +lib-version+ "-0"))
 
 (task-options!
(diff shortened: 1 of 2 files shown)


Deraen:
Alternative to downloading build file from cdn would be to build the file from repo. React-motion readme mentions `npm run prerelease`.

There is one wexample of calling npm from Boot: https://github.com/cljsjs/packages/blob/d1cbb8a9be1d6737028ede59522323f2a4585ca8/blockapps/build.boot

> On a side note I had to include clj-http dependency, missing dependency of the boot-cljsjs task?

Download task loads clj-http into a pod,  so a dependency should not be required: https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L83

edannenberg:
> Alternative to downloading build file from cdn would be to build the file from repo. React-motion readme mentions npm run prerelease.

I was trying to avoid that as it is not really portable in my mind, but if that is a acceptable solution I will change the PR.

>> On a side note I had to include clj-http dependency, missing dependency of the boot-cljsjs task?

> Download task loads clj-http into a pod, so a dependency should not be required: https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L83

I could reproduce the issue by setting `BOOT_HOME` to a fresh location on another drive, `boot package` fails with:

    Downloading react-motion.js
    clojure.lang.ExceptionInfo: java.lang.NoClassDefFoundError: IllegalName: compile__stub.clj_http.headers.clj-http.headers/HeaderMap, compiling:(clj_http/headers.clj:105:1)

Adding `clj-http` dependency to `build.boot` fixes the error. Using the default `BOOT_HOME` I could not reproduce the issue. Maybe just one more of those lovely Windows quirks.

Any ideas on the minify issue?

edannenberg:
As the minify task runs fine on CI it's apparently another Windows issue on my rig, running everything via gitbash. Anyways, looking good now.

PR #389, opened 2016-01-31
Upgrade ol3 to latest version

tatut:
I simply changed the version numbers and checksums. Tried locally with figwheel and :advanced build, both seemed to work fine.


Deraen:
CI had some OOM problems probably caused by large archive files but I think I got that fixed.

PR #393, opened 2016-02-02
add d3-cloud package

pbaille:
This is my first attempt to create a package, it works for me locally, the package task is a little goofy so feel free to refactor it

### DIFF
--- a/d3-cloud/resources/cljsjs/d3-cloud/common/d3-cloud.ext.js
+++ b/d3-cloud/resources/cljsjs/d3-cloud/common/d3-cloud.ext.js
@@ -1,0 +1,2 @@
+var d3 = {};
+d3.layout.cloud = function () {};
--- a/d3-cloud/resources/cljsjs/d3-cloud/common/d3-cloud.ext.js
+++ b/d3-cloud/resources/cljsjs/d3-cloud/common/d3-cloud.ext.js
@@ -1,2 +1,19 @@
 var d3 = {};
 d3.layout.cloud = function () {};
+d3.layout.cloud.prototype = {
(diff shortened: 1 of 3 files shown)


Deraen:
Looks alright.

A problem with urls is that they refer to master branch instead of specific release. Preferred way would be to download `https://github.com/jasondavies/d3-cloud/archive/v1.2.1.zip` and sift file from `d3-cloud-1.2.1/build/d3.layout.cloud.js`.

Also, minify task can be used to create minified version for production folder.

Deraen:
Okay package task looks good now.

How well have you tested the extern? Looking at the readme, looks like `cloud` instance has some instance methods: https://github.com/jasondavies/d3-cloud#cloud

I think there should be defined in the extern, something like:

```js
d3.layout.cloud.prototype.on = function(type, listener) {};
```

d3-cloud/resources/cljsjs/d3-cloud/common/d3-cloud.ext.js
```diff
@@ -0,0 +1,2 @@
+var d3 = {};
+d3.layout.cloud = function () {};
```
Deraen: This might fail because `d3.layout` is null. Probably it should be defined as empty object.

pbaille: it work's locally here
I don't really know how externs works, but d3 is in dependencies so maybe it makes it work

Deraen: Okay.

PR #401, opened 2016-02-05
[exif] add exif.js library

upgradingdave:
This is my first pull request to cljsjs, so if I missed anything or if I can change anything to be more helpful, please let me know. 

Tested in simple and advanced compilation. Here's an example of a widget using this library: http://www.upgradingdave.com/blog/html/exif.html

Thanks, 
Dave

### DIFF
--- a/exif/README.md
+++ b/exif/README.md
@@ -1,0 +1,31 @@
+# cljsjs/exif
+
+[](dependency)
+```clojure
+[cljsjs/exif "2.1.1-1"]
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs]
+feature of the Clojurescript compiler. After adding the above
+dependency to your project you can require the packaged library like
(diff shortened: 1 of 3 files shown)


Deraen:
Looking at the implementation, I think that the extern file is not complete?

I think following properties and methods should be declared in extern: `EXIF.Tags`, `EXIF.TiffTags`, `EXIF.GPSTags`, `EXIF.StringValues`, `EXIF.getTag`, `EXIT.getAllTags`, `EXIF.readFromBinaryFile`. (I did a quick search for `EXIF.` in the file).

upgradingdave:
Good call, I only put in what I needed, I'll add those to the ext as soon as I have a chance, thanks!

upgradingdave:
I updated the externs file, please let me know if I can clean anything else up. Thanks again.

Deraen:
Looks good. Thanks for contribution!

PR #403, opened 2016-02-08
[moment] Add moment.relativeTimeThreshold extern

actionshrimp:
Added extern for this function:
http://momentjs.com/docs/#/customization/relative-time-threshold/


crisptrutski:
LGTM

PR #406, opened 2016-02-09
[react-bootstrap] Add missing navbar externs

edpaget:
Just includes some components that were not in the externs file


Deraen:
If you want this deployed to Clojars you should also update build number in `build.boot`.

edpaget:
Done. Thanks for looking at it.

PR #420, opened 2016-02-16
[syn] Add syn

retro:
Adds [syn](http://github.com/bitovi/syn) - Standalone Synthetic Event Library

### DIFF
--- a/syn/README.md
+++ b/syn/README.md
@@ -1,0 +1,19 @@
+# cljsjs/syn
+[](dependency)
+```clojure
+[cljsjs/syn "0.1.2"] ;; latest release
+```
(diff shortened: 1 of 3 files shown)


Deraen:
As you are using `deps-cljs` task you should not also have `deps.cljs` file, as it will be overwritten by the task.

Or do I miss something?

retro:
@Deraen Ok, is it a better practice to remove the `deps-cljs` task or the `deps.cljs` file?

Deraen:
To remove `deps.cljs` and use the task.

retro:
`deps.cljs` file is removed

PR #422, opened 2016-02-18
[chartist] Fix chartist Line extern

robert-stuttaford:
The API is `Line`; the extern has `Linex`. Likely a typo.

### DIFF
--- a/chartist/build.boot
+++ b/chartist/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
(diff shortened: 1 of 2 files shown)


Deraen:
Could you also update the build number so that I only need to merge PR to make the release?

robert-stuttaford:
Gosh, of course. Doing so now!

robert-stuttaford:
Done, @Deraen – thank you for your speedy response!

Deraen:
I referred to just increasing the build number (`-1`), updating the library in the process is good but now the checksums don't match.

PR #427, opened 2016-02-24
[jquery] Add modal function

zjhmale:
add modal function 

```clojure
(.modal (js/$ "#id") "hide")
```

@Deraen

### DIFF
--- a/jquery/resources/cljsjs/common/jquery.ext.js
+++ b/jquery/resources/cljsjs/common/jquery.ext.js
@@ -252,6 +252,12 @@ jQuery.prototype.appendTo = function(target) {};
 jQuery.prototype.attr = function(arg1, arg2) {};
 
 /**
+ * @param {string} arg1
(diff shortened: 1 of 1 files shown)


Deraen:
Does jQuery really have modal function?

zjhmale:
raw jquery does not have modal function, but if you work with bootstrap modal windows, you have to call modal function with a jquery object.

Deraen:
Those functions should be defined on bootstrap extern, not on jquery extern.

Deraen:
Also, it has already been defined on bootstrap extern: https://github.com/cljsjs/packages/blob/master/bootstrap/resources/cljsjs/common/bootstrap.ext.js#L32

zjhmale:
my bad...  just find it [here](https://github.com/cljsjs/packages/blob/master/bootstrap/resources/cljsjs/common/bootstrap.ext.js#L32)

PR #428, opened 2016-02-24
[react-motion] Remove a mistyped character

hatemogi:
There is a mistyped character in react-motion.ext.js.


Deraen:
Please update build identifier in build.boot if you want this deployed.

hatemogi:
No problem. Thank you for maintaining these packages.

PR #432, opened 2016-02-27
Update react-select beta6 -> beta10

sp3ctum:
As you know I already submitted this patch, but it was based on an old version of the repo. This one should be based on the bleeding edge.


Deraen:
Deployed. Thanks.

PR #433, opened 2016-02-27
Update README.md

kblw:
bad version


Deraen:
Yeah this happens currently probably quite often as current Boot tasks don't update this automatically. I'm inclined to replace version numbers of readmes with Clojars badges.

PR #440, opened 2016-03-02
[codemirror] Add mode/meta.js

djwhitt:
meta.js lives in the root of the mode directory rather than a
subdirectory, so it is not picked up by the regular mode regex. It is
not itself a mode - it contains function to map extensions, mime types,
and names to modes.  However, since it is closely related to modes and
it seems unlikely there will ever be a "meta" mode I think it makes
sense to place it under cljsjs.codemirror.mode in the namespace
hierarchy. I'm open to suggestions on the placement though if anyone
has other suggestions.


Deraen:
As the file is under `mode` directory, I think this makes sense.

PR #443, opened 2016-03-04
[react-mdl] Fixed missing main project module. Moved extras to their own package.

Shearerbeard:
Fixed issue where main project module was being exported.
Bumped ReactMDL to the newest version.
Moved dependencies from the "extra" directory to their own package `react-mdl-extra` to avoid an issue with the dependencies being orphaned.
Added proper externs file for the main exported module.


Deraen:
I'm not sure I understand the need for separate extra package.

I think it generally makes sense to package files from single source to single package.

Seems especially strange that the core namespace depends on the extra namespace? But core package doesn't depend on extra package.

Shearerbeard:
This library is unique in that it contains a modified version of the material design light js and css that the author has modified to work with React. The extra/material.css and extra/material.js have to be loaded before ReactMDL.js. I didn't find any success with getting them to load properly bundled in the same cljsjs package.

Deraen:
You can get them to load before core code by creating proper `deps.cljs`, but you can't do that with `deps-cljs` task. deps.cljs can define multiple namespaces and depenendencies between them.

Shearerbeard:
Understood - I thought the convention for this repository was to stick with the boot.cljsjs auto packaging and deps generation. I would be happy to consolidate if thats possible.

Deraen:
Creating `deps.cljs` manually is fine when `deps-cljs` is not enough.

Shearerbeard:
I'll change that and get it committed - the boot script wont overwrite the deps.cljs?

Deraen:
Just remove `deps-cljs` task from pipeline and nothing will touch `deps.cljs`.

Shearerbeard:
Updated to consolidated cljsjs package.

PR #450, opened 2016-03-06
[react-mdl] moving css assets to public/assets

Shearerbeard:
Changed bathing for css assets in jar to move to a public directory for client side projects.

### DIFF
--- a/react-mdl/build.boot
+++ b/react-mdl/build.boot
@@ -10,7 +10,7 @@
          '[boot.task-helpers :as helpers])
 
 (def +lib-version+ "1.4.3")
-(def +version+ (str +lib-version+ "-0"))
+(def +version+ (str +lib-version+ "-1"))
 
 (task-options!
  pom  {:project     'cljsjs/react-mdl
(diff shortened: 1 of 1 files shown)


martinklepsch:
hey @Shearerbeard, thanks for your contribution. Generally we want to keep things under a predictable namespace/directory (i.e. `cljsjs/`). Users then need to move files to the places they need them or setup handlers to serve them on their own. There's a bit of documentation about this in the wiki, if you're using Boot this will be easy :smile: https://github.com/cljsjs/packages/wiki/Non-JS-Assets

A change like this may seem handy but its impossible to anticipate all the setups people will have and thus consistency/predictability is the better option IMHO.

Let me know if you have further questions. :)

Deraen:
Just created https://github.com/Deraen/ring-cljsjs (will be available in Clojars once I get GPG to co-operate).

Shearerbeard:
Awesome Juho, this is a great solution to serve other static assets from cljsjs packages.

Deraen:
Available now https://clojars.org/ring-cljsjs

PR #451, opened 2016-03-06
[react-grid-layout] bump to version 0.10.8

currentoor:
There was one part I was not completely sure about. Grid elements can have a magic prop, called `_grid`, that gets used by ReactGridLayout. So we don't want it to get compiled away in our source.
https://github.com/STRML/react-grid-layout#usage

So I put it on `React.ReactAttribute`, just like the `key` prop.
https://github.com/cljsjs/packages/blob/master/react/resources/cljsjs/react/common/react.ext.js#L282

I'm not sure if this is the correct way to do it.

### DIFF
--- a/react-grid-layout/README.md
+++ b/react-grid-layout/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-grid-layout "0.8.5-0"] ;; latest release
+[cljsjs/react-grid-layout "0.10.8-0"] ;; latest release
 ```
 [](/dependency)
 
--- a/react-grid-layout/build.boot
+++ b/react-grid-layout/build.boot
@@ -1,12 +1,12 @@
+(def +lib-version+ "0.10.8")
+(def +version+ (str +lib-version+ "-0"))
+
 (set-env!
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.0"  :scope "test"]
                   [cljsjs/react "0.14.3-0"]])
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
-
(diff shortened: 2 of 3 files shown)


Deraen:
Is `_grid` used by ClojureScript code? Extern only affect the calls from ClojureScript to external libraries. Not calls inside or between external libs.

currentoor:
@Deraen, yes. There are two ways to use the grid. Put the layout information on the grid element.
```cljs

(js/React.createElement
 js/ReactGridLayout
 (clj->js {:className       "layout"
           :onLayoutChange  track-layout
           :layout          [{:i "1" :x 1 :y 1}
                             {:i "2" :x 0 :y 1}
                             {:i "3" :x 1 :y 1}]
           :verticalCompact false
           :cols            24
           :width           700
           :rowHeight       30})
 (dom/div (clj->js {:key "1"}))
 (dom/div (clj->js {:key "2"}))
 (dom/div (clj->js {:key "3"})))
```
Or you can put the layout information directly on the grid element's children.
```cljs
(js/React.createElement
 js/ReactGridLayout
 (clj->js {:className       "layout"
           :onLayoutChange  track-layout
           :verticalCompact false
           :cols            24
           :width           700
           :rowHeight       30})
 (dom/div (clj->js {:key "1" :_grid {:i "1" :x 1 :y 1}})) ; <- using _grid in cljs
 (dom/div (clj->js {:key "2" :_grid {:i "2" :x 0 :y 1}}))
 (dom/div (clj->js {:key "3" :_grid {:i "3" :x 1 :y 1}})))
```

Deraen:
I don't think extern is required is this case. The `_grid` is a property of Object value here. Extern is needed when calling methods, e.g. `(._grid foobar)`. Anyway, there is no (much) harm in having additional externs.

PR #460, opened 2016-03-10
react 0.14.5

ashercoren:


Deraen:
Packaging changes look good but I guess you have completely replaced hand written extern with generated once based on huge diff?

Any special reason for that? I would guess API doesn't differ much (or probably at all) between minor versions. I would prefer keeping the extern as is.

ashercoren:
@Deraen You're right - the API didn't change. I fixed the pull request to use the original extern.

Deraen:
Thanks. I'll try to get this merged soon. I'm thinking of implementing some CI changes at the same time to get with-addons deployed by CI, currently only normal package is deployed automatically.

Deraen:
The changes were quite extensive so I reimplemented this myself with the new packaging.

PR #462, opened 2016-03-11
update react-input-mask to 0.5.10 and minify it

piranha:


react-input-mask/build.boot
```diff
@@ -23,11 +23,14 @@
 
 (deftask package []
   (comp
-    (download :url "http://sanniassin.github.io/react-input-mask/javascripts/InputElement.js"
-	      :checksum "CF24D3069E1A47DDAA5A708BF4E61D9E")
+    (download :url "https://raw.githubusercontent.com/sanniassin/react-input-mask/2489e7eeb9ba744e68046456c62febd8cc77802c/InputElement.js"
```
Deraen: Would be preferred to refer correct version using `+lib-version+` instead, but unfortunately looks like the project doesn't have any git tags. I think correct commit for the 0.5.10 release is: https://github.com/sanniassin/react-input-mask/commit/05c101e4937d54966f673113a1432afb53fb9541

piranha: Right, I just thought to use latest version. :) Okay, I'll fix that. :)

piranha:
Fixed.

PR #464, opened 2016-03-14
Add react-autosuggest

cdorrat:
This pull requests adds the react-autosuggest library from http://react-autosuggest.js.org/


Deraen:
The packaging files look okay but could you squash the commits into one? 11 commits is a bit much :)

`git rebase -i HEAD~11` and changing "pick" to "squash" for all but the first line should work. Then force push or new branch & new pr.

PR #468, opened 2016-03-14
Added react-autosuggest

cdorrat:
This is the squashed version the pull request that adds the react-autosuggest library from http://react-autosuggest.js.org/


Deraen:
Thanks

PR #478, opened 2016-03-19
[react-relay] Add Relay package

DomKM:


Deraen:
Thanks.

Using webjar for source is nice trick, takes care of caching etc. :)

PR #489, opened 2016-03-28
Fix google-diff-match-patch production build

akjetma:
The distributed minified version of the library is missing a semi-colon that causes a cryptic error message when you use more than one cljsjs package in a project, didn't notice until now.

### DIFF
--- a/google-diff-match-patch/README.md
+++ b/google-diff-match-patch/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/google-diff-match-patch "20121119-0"] ;; latest release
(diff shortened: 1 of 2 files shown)


Deraen:
If you want to have minified file, alternatives would be to use [minify task](https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L134) from boot-cljsjs or create a task to add the missing `;` to the end of file. Something like [replace-content task](https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L168), but not using regex.

PR #495, opened 2016-03-31
add google platform library (for authentication)

yeehaa123:
This is the library that you need to use google's oauth2 authentication. There are some quirks. It seems that Google wants to force you to use their CDN and does not have a public repo or github repo for this library. Therefore the version that I mention is bogus (google does not externally version this) and it may break unexpectedly. When this occurs, an easy fix is described in the README

### DIFF
--- a/platform/README.md
+++ b/platform/README.md
@@ -1,0 +1,27 @@
+# cljsjs/platform
+
+[](dependency)
+```clojure
+[cljsjs/platform "1.0.0-0"] ;; latest release
+```
+[](/dependency)
+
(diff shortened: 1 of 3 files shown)


yeehaa123:
Okay, checksum is already different on CI. This is the problem that I have mentioned above. Anyone knows where to find a properly versioned repo for this lib?

Deraen:
I think it is best to not packages files the vendor thinks should be used from CDN. You could inlcude only the extern file in this package and mention in the readme that user has to add script tag to html to include CDN file.

Deraen:
I packaged this as extern only package. As I'm not sure what is the proper name for the library and there seems to be several Platform.js libraries, I named this `google-platformjs-extern` to avoid confusion.

PR #502, opened 2016-04-06
cljsjs.tabletop init commit

leontalbot:


Deraen:
The extern file seems to contain complete source code. It should only contain the function declarations: https://github.com/cljsjs/packages/blob/master/react/resources/cljsjs/react/common/react.ext.js

https://github.com/cljsjs/packages/wiki/Creating-Externs

PR #515, opened 2016-04-13
adds chroma

levitanong:


Deraen:
Please download the file using specific version (tag): https://raw.githubusercontent.com/gka/chroma.js/v1.1.1/chroma.js

And preferable the package should include the minified file also. One seems to be available in the repo and you could probably load the zip from github to get the both files at once.

levitanong:
@Deraen apologies! Got lazy and just followed what I did for the other stuff I packaged that didn't have proper tags. Fixed!

PR #519, opened 2016-04-16
[ical] add ical.js library

upgradingdave:
I'd like to contribute this just in case anyone else would like to use ical.js from clojurescript. 

Please let me know if you notice anything I can clean up or change, thanks!


ical/README.md
```diff
@@ -0,0 +1,46 @@
...
+so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.ical))
+```
+
+Here's a simple example of how to parse the contents of an `.ics` file:
+
+```
```
Deraen: Very minor style issue here: extra newline and missing clojure filetype from code block.

Looks good otherwise.

Deraen:
I merged this as the only minor remarks where on readme, it can be fixed later but it doesn't matter much.

Issue #518, opened 2016-04-15
Version bump: d3 3.5.16


devurandom:
It would be nice if the latest bugfix release of D3js could be packaged. cljsjs/d3 is currently at 3.5.7, while 3.5.16 is the current upstream version.

ducky427:
#523 addresses this.

PR #523, opened 2016-04-18
[d3] Updated to v3.5.16

ducky427:


ducky427:
This should address #518

PR #525, opened 2016-04-19
Add Photoswipe library

AdamFrey:
http://photoswipe.com

Photoswipe splits its default UI javascript into a separate file, so here I'm generating a second entry into the `:foreign-libs` key of `deps.cljs` for `cljsjs.photoswipe-ui-default`.

### DIFF
--- a/photoswipe/README.md
+++ b/photoswipe/README.md
@@ -1,0 +1,21 @@
+# cljsjs/photoswipe
+
+[](dependency)
+```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
The tests were failing because this branch was based on old commit. Not super dangerous but basing the changes on the latest commit prevents problems such as this and makes it faster for me to merge PRs.

AdamFrey:
good to know. Thanks!

PR #527, opened 2016-04-20
added material-ui

madvas:
http://www.material-ui.com/

### DIFF
--- a/material-ui/README.md
+++ b/material-ui/README.md
@@ -1,0 +1,20 @@
+# cljsjs/material-ui
+
+http://www.material-ui.com/#/
+
+[](dependency)
+```clojure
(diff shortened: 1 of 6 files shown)


Deraen:
Hmm, I don't like packaging files from random repositories. Am I correct in assuming that your bundle exists because of https://github.com/callemall/material-ui/issues/262?

How is the bundle created? Could we automate the bundle creation here so we could load the source from original repo?

madvas:
Yes, there's a issue that they don't have bundled version and also that it includes react, which is not desired. I create bundle in my other repo, but okay I'll try to automate the process here.

Deraen:
Looks good now, thanks!

PR #530, opened 2016-04-23
[leaflet] Add some missing leaflet externs

AdamFrey:
Externs created from documentation: http://leafletjs.com/reference.html


Deraen:
Thanks!

PR #532, opened 2016-04-26
pinterest JS sdk externs

hoopes:
I'm pretty much dipping my toe in the world of github, so if there's something wrong here, please let me know, and I'll remedy asap.

I pretty much copied the facebook package, since the two are so similar. I wasn't certain about the version number, so i used today's date, since that's the sdk that exists today.

### DIFF
--- a/pinterest/README.md
+++ b/pinterest/README.md
@@ -1,0 +1,15 @@
+# cljsjs/pinterest
+
+[](dependency)
+```clojure
+[cljsjs/pinterest "v20160425-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler.
+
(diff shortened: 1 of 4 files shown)


AdamFrey:
Thanks!

One more file you need to add is a `deps.cljs` in the `resources` directory. [Here's the example](https://github.com/cljsjs/packages/blob/master/facebook/resources/deps.cljs) from the Facebook package. `deps.cljs` are necessary for externs only packages.

hoopes:
Added - my directory structure is slightly different than the facebook one - `facebook/resources/cljsjs/common/facebook_javascript_sdk.ext.js` vs `pinterest/resources/cljsjs/pinterest/common/pinterest.ext.js` (mine has the extra pinterest directory, but I thought that's what the readme asked for)

In any case, let me know if it's good to go. Thanks!

AdamFrey:
yep, that directory structure is preferred. Thanks for your contribution!

PR #535, opened 2016-04-26
Updated version and added react to the build

madvas:
Currently, material-ui has to be built together with react, otherwise [react-tap-event-plugin](https://github.com/zilverline/react-tap-event-plugin) doesn't work. So this build includes `react` `react.dom` `react-tap-event-plugin` `material-ui`. This should not be needed in a future.

### DIFF
--- a/material-ui/README.md
+++ b/material-ui/README.md
@@ -4,9 +4,19 @@ http://www.material-ui.com/#/
 
 [](dependency)
 ```clojure
-[cljsjs/material-ui "0.14.4"] ;; latest release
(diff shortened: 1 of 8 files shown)


Deraen:
Wasn't the purpose of webpack configuration to be able to build standalone file?

madvas:
Yes, it was. But I've tried many times I couldn't get react-tap-event-plugin working anyhow. (http://www.material-ui.com/v0.15.0-beta.2/#/get-started/installation) . It works only when it's built together, probably because it includes some parts of react, which are not normally included. Probably related to this http://stackoverflow.com/questions/29881439/react-tap-events-and-material-ui

PR #536, opened 2016-04-28
[react-bootstrap] update to 0.29.2 with react 15

smee:
I used the [recommended externs generator](http://jmmk.github.io/javascript-externs-generator/), which seems to identify all elements that had to be inserted manually before.

### DIFF
--- a/react-bootstrap/README.md
+++ b/react-bootstrap/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-bootstrap "0.28.1-1"] ;; latest release
+[cljsjs/react-bootstrap "0.29.2-0"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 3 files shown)


AdamFrey:
I'm not sure that the externs you added are necessary, because they I believe most of them should be covered by the `react` package externs. Then again it might not hurt anything to include them if they were auto-generated.

smee:
Oh, sorry, did not see that. I had to load react.js as well as react-bootstrap.js to get the externs generator to work. I didn't see that it also generated declarations for react. Should I remove them? In my application everything works (I use this new version of react-bootstrap and cljsjs/react-dom "15.0.1-1").

AdamFrey:
quoting @Deraen "The externs shoud be fine even if they include unncessary stuff. It might cause minor size increase in optimized output file, but the effect is probably minimal."

especially because it looks like the generated externs include some important ones that were missing, e.g. https://github.com/smee/packages/blob/master/react-bootstrap/resources/cljsjs/react-bootstrap/common/react-bootstrap.ext.js#L817

Deraen:
In this case many of the unnecessary methods were duplicates of basic React methods. As those method calls are already  excluded from the name mangling by React extern, including these doesn't have any effect on the output. One theoretical effect could be that Closure optimizer might get slower the more externs there are, but this is pure guess.

PR #546, opened 2016-05-02
Add rangy-core

oakes:


oakes:
Previously, the test failed because I had two separate packages in here, on which relied on the other. I temporarily removed the second on and will submit it later in a separate PR.

PR #553, opened 2016-05-03
Add react-infinite package

smitch88:
#### What

Adds `react-infinite` version tag 0.9.0 w/ externs, boot file, and README

https://github.com/seatgeek/react-infinite

**Notes:** Auto generated externs


Deraen:
Thanks! Nice docs on the extern file.

PR #561, opened 2016-05-06
Adds bcrypt library version 2.3.0

smitch88:
#### What

Adds `bcrypt` version 2.3.0 the latest tagged release.

Notes: The provided externs in the library itself seemed to be out of date. I updated it via a hand rolling but was tested with optimizations `:none` and `:advanced`. It exposes the lib thru js/dcodeIO so I added an example in the README.md about that.


AdamFrey:
thanks

PR #570, opened 2016-05-10
Increased version to 1.5.4

taylorSando:


AdamFrey:
The version number in the README should be `1.5.4-0`

Deraen:
Thanks.

Updating readme manually is not very important, automated script will take care of that if necessary.

PR #571, opened 2016-05-10
[filestack] inital

martinklepsch:
This adds [Filestack](https://www.filestack.com/docs/).

Unfortunately they don't version their builds properly so I just used 0.0.1 as version & added a little note about the lack of versioning to the readme.

### DIFF
--- a/filestack/README.md
+++ b/filestack/README.md
@@ -1,0 +1,22 @@
+# cljsjs/filestack
+
+[](dependency)
+```clojure
+[cljsjs/filestack "0.0.1-0"] ;; latest release
+```
+[](/dependency)
+
+**Note:** Filestack does not publish versioned artifacts, for that reason `0.0.1` is used as placeholder.
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.filestack))
+```
+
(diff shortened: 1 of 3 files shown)


Deraen:
If they presume users load the code from their cdn, perhaps this counts as a special case where only the extern should be packaged: https://github.com/cljsjs/packages/wiki/Extern-only-packages ?

martinklepsch:
@Deraen I considered that but the md5 is always the same so I don't think they do any special on the flight packaging like e.g. Google Maps.

Deraen:
What about these bower and npm packages mentioned on [Filestack docs](https://www.filestack.com/docs/)?

https://www.npmjs.com/package/filepicker-js
https://github.com/filepicker/filepicker-js-bower
https://github.com/filepicker/filepicker-js

filepicker.io seems to redirect to filestack.io.

martinklepsch:
@Deraen uh, this looks good. I'll update the PR to get a versioned artifact from there. Still I'm not seeing why this should be an externs only package, do you?

Deraen:
No reason if there is a proper versioned artifact.

If only a file from a vender cdn is available updates would be hard to manage. Even if the vendor doesn't do "on the flight packaging like Google Maps" they might update the file in two weeks or two months. Without versioning this is hard to notice and hard to communicate to library users.

martinklepsch:
> Without versioning this is hard to notice and hard to communicate to library users.

Agreed, will update the package to use versioned artifacts. I'll recheck with Filestack if there are any other reasons why you should inline their js file.

martinklepsch:
Heard back from Filestack, there's no reason to not inline it.

PR #580, opened 2016-05-14
Added the async library

Risto-Stevcev:


AdamFrey:
It's such a bold move to name a library simply "async" 😮

PR #583, opened 2016-05-16
[leaflet-locate] Add leaflet.locate support

werenall:


Deraen:
I think "leaflet-locatecontrol" would be better name. It would be more in line with names in other package systems etc: https://www.mapbox.com/mapbox.js/plugins/#leaflet-locatecontrol https://www.npmjs.com/package/leaflet.locatecontrol

As this library depends on Leaflet, it should have dependency on cljsjs/leaflet package: https://github.com/cljsjs/packages/blob/master/react-leaflet/build.boot#L4 and mark the provided foreign-dependency as depending on cljsjs.leafet: https://github.com/cljsjs/packages/blob/master/react-leaflet/build.boot#L30

How did you generate the extern file? It looks like the file includes the whole Leaflet externs in addition to Leaflet.Locate externs. It is unnecessary to include base leaflet stuff in the extern. I think you could either  try using "L.Control" as JS Object name in the generator (not sure if this works) or just manually remove stuff already included in Leaflet extern (lines 2-2199).

BTW. leaflet-draw seems to also miss the dependencies and has unnecessary stuff on the extern, would be super great if you can fix that package also!

Issue #510, opened 2016-04-10
Remove JQuery dependency for Highcharts


vincent-dm:
The Highcharts library used to depend on JQuery, but this is not the case anymore ([in 2013 they added a standalone framework](http://www.highcharts.com/articles/2-news/58-highcharts-standalone-framework) as an alternative to JQuery, but since Dec 2015 even that is not needed anymore).

The old (JQuery) way to create a chart in cljs:
```clojure
(.highcharts (js/$ (.getElementById js/document "id")) (clj->js chart-config)))
```

However, when the standard Highcharts library is loaded in the browser, the Highcharts.Chart method can attach a chart into the DOM without using JQuery.
An example:
```clojure
(js/Highcharts.Chart. (.getElementById js/document "id") (clj->js chart-config))
```

Since JQuery is a pretty hefty dependency, it would be good to remove it. I'm aware that most examples of Highcharts still use JQuery syntax, but since cljs-users already have to re-write every code example, I don't see why we should keep relying on JQuery.

Thanks!

vincent-dm:
@martinklepsch do you think we can drop jquery?

ducky427:
@vincent-dm, would excluding `cljsjs/jquery` from your dependency list in `project.clj` work? See [this](https://github.com/technomancy/leiningen/blob/master/sample.project.clj#L48) for more details on how to do it.

(admitting that a new release would be a better option)

yatesco:
I didn't realise this so my PR to update to HC4.2.5 (https://github.com/cljsjs/packages/issues/585) still lists the jquery dependency.

Deraen:
Merged in #586

vincent-dm:
Thanks!

Issue #585, opened 2016-05-16
Upgrade Highcharts to 4.2.5


yatesco:
Deleted and issued new PR

yatesco:
Fixed in https://github.com/cljsjs/packages/pull/586

PR #586, opened 2016-05-16
Highcharts - upgrade to 2.4.5 and removed jquery

yatesco:


yatesco:
Fixes https://github.com/cljsjs/packages/issues/510 and https://github.com/cljsjs/packages/issues/585

Deraen:
Did you try generating new externs? One note in the change log looks like it might change the API:

- Added new constructor, Highcharts.chart, that allows instanciating a chart without the new keyword.

yatesco:
I'm new to this and boot so I might be mistaken, but target/common/highcharts.ext.js contains:
```
var Highcharts = {
...
  "Chart": function () {},
  "chart": function () {},
...
}
```

which  I think is sufficient, but no, I didn't manually edit that.

Deraen:
Yeah, that looks sufficient.

P.S. I added a few notes about updating packages to wiki: https://github.com/cljsjs/packages/wiki/Updating-packages

yatesco:
Great - thanks!

PR #587, opened 2016-05-17
[leaflet-draw] dependencies added and .ext.js reduced

werenall:


leaflet-draw/build.boot
```diff
@@ -1,34 +1,36 @@
...
+        :scm         {:url "https://github.com/cljsjs/packages"}
+        :license     {"MIT" "http://opensource.org/licenses/MIT"}})
 
 (require '[clojure.java.io :as io])
 
 (deftask package []
   (comp
-    (download :url      (str "https://github.com/Leaflet/Leaflet.draw/archive/master.zip")
-              :checksum "C2316EBEDD093B67F5F7083F75097444"
+    (download :url      (str "https://github.com/Leaflet/Leaflet.draw/archive/0.2.3.zip")
```
Deraen: It is a best practice to use `format` to create the url using `+lib-version+` here, like this: https://github.com/cljsjs/packages/blob/master/airbrake/build.boot#L28

PR #589, opened 2016-05-17
[material-ui] Fix to provide react from built file

### LINKED ISSUE #538: material-ui can't find ReactDOMServer
screen shot 2016-04-28 at 8 51 50 am

Latest version of [material-ui](https://github.com/cljsjs/packages/tree/master/material-ui) is unable to find ReactDOMServer with [reagent](https://github.com/reagent-project/reagent).

Steps to reproduce this issue:
- `lein new figwheel hello-world -- --reagent`
- add `[cljsjs/material-ui "0.15.0-beta.2-10"]` to project.clj 
- change project.clj reagent dep to use `[reagent "0.6.0-alpha" :exclusions [cljsjs/react cljsjs/react-dom]]`
- add `cljsjs.material-ui` and `cljsjs.material-ui-svg-icons` to ns :require in core.cljs
- `cd hello-world/`
- `lein figwheel`
- then open http://localhost:3449/ in a browser, and check console.

edpaget:
Make sure the file that provides cljsjs/react is named as the provider,
to prevent problems with how ClojureScript orders dependencies.

Closes #538

### DIFF
--- a/material-ui/build.boot
+++ b/material-ui/build.boot
@@ -9,7 +9,7 @@
          '[boot.util :refer [sh]])
 
 (def +lib-version+ "0.15.0")
-(def +version+ (str +lib-version+ "-0"))
+(def +version+ (str +lib-version+ "-1"))
 (def +lib-folder+ (format "material-ui-%s" +lib-version+))
 
 (task-options!
@@ -50,7 +50,7 @@
                 (binding [boot.util/*sh-dir* (str (io/file tmp +lib-folder+))]
                          (do ((sh "npm" "install"))
                              ((sh "npm" "install" "webpack"))
-                             ((sh "npm" "run" "build"))
+                             ((sh "npm" "install" "babel-loader"))
                              ((sh "./node_modules/.bin/webpack"))
                              ((sh "./node_modules/.bin/webpack" "--production"))
                              ((sh "./node_modules/.bin/webpack" "--svg-icons"))
--- a/material-ui/resources/deps.cljs
+++ b/material-ui/resources/deps.cljs
@@ -1,9 +1,5 @@
-{:foreign-libs [{:file     "cljsjs/react/development/react.inc.js",
-                 :provides ["cljsjs.react"],
-                 :file-min "cljsjs/react/production/react.min.inc.js"}
-                {:file     "cljsjs/material-ui/development/material-ui.inc.js",
-                 :provides ["cljsjs.material-ui"],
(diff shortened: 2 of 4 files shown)


edpaget:
The problem I'm experiencing using the material-ui package is that I can't control the order ClojureScript decideds to put dependencies in in the cljs_dep.js file it generates. Something is causing it to include the blank React file first, then try to include ReactDOM, which fails (since React is not defined). 

To me it makes sense to just have the file that actually has the React symbol be marked as the provider for cljsjs.react. Feel free to close if this is wrong.

edpaget:
Also found on my local machine (64-bit Linux running Node 6.1.0), the build would fail for Material UI with a a call stack exceeded error while compiling the svg icons. I changed the process to use the babel-loader for webpack instead of the build scripts for Material UI. It should do the same thing, but again happy to remove if we'd prefer to use the Material UI process.

Deraen:
@madvas Can you check and test this? I'm not familiar with material-ui.

madvas:
@edpaget 
`cljsjs/material-ui` has `cljsjs/react` in `provides`
https://github.com/cljsjs/packages/blob/master/material-ui/resources/deps.cljs

And yes, you can have that problem with dependencies, but you can control it. 
Take a look here https://github.com/madvas/cljs-react-material-ui , in instructions there's is comment that you should put `[cljsjs.material-ui]` as a first require in your core file. This should effectively solve your problem. I should add that node to readme here as well.

edpaget:
@madvas It has react as being provided from a blank file. In my project I do have [cljsjs.material-ui] in the main file figwheel is pointed at (I did use you library thanks a bunch for it!), but something is causing ClojureScript to require `om.dom` and its dependencies (including ReactDOM and React), before the Material-UI file.

Basically the problem is if ClojureScript decides the Dependency order is `React -> ReactDOM -> om.dom`, it will load the react.inc.js file which doesn't have anything in it. If we change deps to say react is provided from the `material-ui.inc.js` file (which it is, so I have no idea why you wouldn't want it declared as providing React). Even in this situation. the React symbols will be loaded at the correct time.

madvas:
@edpaget Oh, I get it now, sorry. This just didn't occur to me, don't know why 😆 You're probably right, I will test it and change it as soon as I will find time.

edpaget:
Thanks!

Deraen:
Was this fixed by #591?

edpaget:
Yes

PR #590, opened 2016-05-18
[field-kit] Add Square's field-kit library

jalehman:
[field-kit](http://square.github.io/field-kit/) is a library for validating/masking text inputs.


Deraen:
Thanks

PR #591, opened 2016-05-19
[material-ui] Fixed: deps provides

madvas:
Fixed: Stack size issue


Deraen:
```
./react/resources/cljsjs/react/common/react.ext.js:11: ERROR - constant React assigned a value more than once.
Original definition at ./material-ui/resources/cljsjs/material-ui/common/react.ext.js:11
var React = {};
```

the tests are failing because the extern test is hardcoded to skip react extern in material-ui folder: https://github.com/cljsjs/packages/blob/master/test-extern-files.sh#L8 and the path changed. I suggest you change the line to this to fix the test:
```
    -not -wholename './material-ui/*/react.ext.js'); do
```

PR #594, opened 2016-05-22
[update] Firebase 3.0.*

flyboarder:
Update to Firebase 3.0.0

* Needs externs
* Newer version available from CDN (3.0.2)

Should we get CDN versions instead?


flyboarder:
Relates to:

https://github.com/crisptrutski/matchbox/issues/74

Deraen:
How has the extern been created? It looks quite quite small compared to Firebase API: https://firebase.google.com/docs/reference/js/#firebase.User

martinklepsch:
I opened a ticket with firebase to ask if they could provide externs
directly. Will update here when I hear back.

I think the externs aren't done as they are in the PR, as Matthew
mentioned.

Deraen:
Btw, if you need a release without externs for dev, I think we can do a release with alpha indetifier?

```clj
(def +version+ (str +lib-version+ "-alpha0"))
```

amitburst:
It looks like the Firebase npm package provides externs now.

flyboarder:
@Deraen ready 👍

flyboarder:
@amitburst thanks for the heads up! 😸

PR #599, opened 2016-05-27
[Mithril] Fix version number

jimrthy:
Lein couldn't find in clojars based on existing spec.

Did a search, found this variant, and at least it downloaded.


jimrthy:
I think this is probably just a very minor documentation issue, but this is my first contribution attempt here. So I could be way out in left field.

AdamFrey:
looks good, thanks!

PR #602, opened 2016-05-30
[react-pixi] Add react-pixi 0.8.1

weavejester:
This adds the latest version of @Izzimach's [react-pixi](https://github.com/Izzimach/react-pixi) library to cljsjs.

cljsjs/react-pixi has a dependency on cljsjs/pixi, but not on cljsjs/react, as the library makes several changes to React in order to support Pixi.js, and so includes its own version. This increases the filesize, but appears to have no other effect. I've tested cljsjs/react and a locally installed version of cljsjs/react-pixi together in a few example projects, and they both seem to work fine together.


Deraen:
If this is supposed to be used with the included React, I think this should also contain React externs? And deps.cljs should also provide `cljsjs.react`.

Have you tested this package without cljsjs/react? Global exclusion to cljsjs/react should be the easiest way to test that.

https://github.com/cljsjs/packages/tree/master/material-ui is another package that uses bundled React

As a minor note, webpack.config.js doesn't need to be included in the JAR, it could be moved to either path inside resource-paths which is not included (anything outside `cljsjs/` currently) or to a source-path like `build-resources` (set up by `:source-paths #{"build-resources"}`.

weavejester:
> If this is supposed to be used with the included React, I think this should also contain React externs?

It could do. I thought it would be easier if it didn't, as then it could be plugged in and used without needing to fiddle around with exclusions. But... if other libraries like material-ui already do it (and, I guess, react-with-addons), then maybe it should also contain the react externs.

I'll change it and add a note to the README.

> As a minor note, webpack.config.js doesn't need to be included in the JAR

Ah, I didn't notice it was being included. I'll fix that.

Deraen:
> I thought it would be easier if it didn't, as then it could be plugged in and used without needing to fiddle around with exclusions.

That means that React library is included twice in the result and I think it is based on luck etc. which React version is used?

weavejester:
React would be included twice, but because webpack bundles it in as a local package, there shouldn't be any conflict between the two. That said, it's probably safer to expose it as a extern.

Deraen:
If there is no conflict, I'm fine with either solution.

weavejester:
I'm not having much luck with including `cljsjs.react` in react-pixi and getting it to play nice with other namespaces like `cljsjs.react.dom`. I've also thought about the problem a little bit more, and I'm leaning back toward my initial solution.

My reasoning is that if react-pixi uses its own, local, React, then we don't need to worry about how the patched React interoperates with existing libraries that have a React dependency. We can also update React independently of react-pixi, without worrying whether the update will affect the patched classes. From a usability perspective, it removes the need to add in `:exclusions` for any dependency that depends on cljsjs/react.

It does add around 150KB to the minified file size, but it seems the more reliable solution. I'd rather use the official React to handle the DOM, and have react-pixi work through its own patched copy.

Do you want me to squash my commits?

Deraen:
No need to squash, the commits are clean enough :)

weavejester:
Thanks!

PR #604, opened 2016-06-02
[material-ui] add react-dom and react-dom-server to bundled react

bendlas:
Various packages (like sablono) need `cljsjs.react.dom` along with `js/ReactDOM` and `js/ReactDOMServer`.
This PR adds support for those. It shouldn't make a difference in filesize, as that code is in react anyways.

cc @madvas

### DIFF
--- a/material-ui/README.md
+++ b/material-ui/README.md
@@ -4,7 +4,7 @@ http://www.material-ui.com/#/
 
 [](dependency)
 ```clojure
-[cljsjs/material-ui "0.15.0-3"] ;; latest release
+[cljsjs/material-ui "0.15.0-4"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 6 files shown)


madvas:
This is actually intentional, so it doesn't have to be excluded from libraries like Om/Reagent, which include it. What's your use case? Is something not working for you?

bendlas:
Yeah, I was trying to use rum with `cljsjs/react-dom` excluded aswell. I didn't realize, I could use `cljsjs/react-dom` together with our custom build, since it picks up from `js/React`. Still seems a bit strange to do it this way, since users will have to exclude `cljsjs/react` from Om/Reagent anyway. Also it seems desirable to keep `react-dom` in sync with our build and not pick up a random one no?

madvas:
I think, it's bit more simpler for users, because they have to exclude only react and not other things. But difference is minor. Well, as you said, `react-dom` logic is really included in `react`, react-dom file is only like global variable assignment. So I don't think out of sync is gonna be an issue.

bendlas:
Yeah, I don't really care either way. You're the author.

Issue #271, opened 2015-10-24
Request: HUSL Colors


den1k:
[HUSL](http://www.husl-colors.org/) provides the human friendliest colorspace I know. Would love to have it as a dep in CLJS.

Deraen:
Seeing that the implementation is only 300 lines (https://github.com/husl-colors/husl/blob/master/husl.coffee), you could very easily just re implement it in Clojure. It could be crossplatform for Clojure and ClojureScript and you wouldn't need a wrapper.

den1k:
@Deraen there already is a Java port https://github.com/husl-colors/husl/blob/master/ports%2Fjava%2FHuslConverter.java

Deraen:
Though I recommended re-implementation, a PR to add this package would be accepted.

Deraen:
Merged.

PR #607, opened 2016-06-08
[husl] Add the HUSL.js color library

rafd:
Info about the library here:
 - http://www.husl-colors.org/
 - https://github.com/husl-colors/husl

### DIFF
--- a/husl/README.md
+++ b/husl/README.md
@@ -1,0 +1,18 @@
+# cljsjs/husl
+
+[](dependency)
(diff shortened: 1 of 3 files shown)


rafd:
Addresses https://github.com/cljsjs/packages/issues/271

rafd:
I have tested it with a project locally both in dev and production.

Deraen:
Thanks!

PR #609, opened 2016-06-11
upgrade to latest upstream version, add react requirement

euccastro:
Without the cljsjs.react requirement, the library used to fail to load for me, even if I added it to my project's `:dependencies` separately.

### DIFF
--- a/victory/build.boot
+++ b/victory/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "0.6.0")
+(def +lib-version+ "0.8.0")
 (def +version+ (str +lib-version+ "-0"))
 
 (task-options!
@@ -18,11 +18,12 @@
 (deftask package []
   (comp
     (download :url (format "https://github.com/FormidableLabs/victory/archive/v%s.zip" +lib-version+)
-              :checksum "EDE1D4CED3A156E51E6C762DAA5F5815"
+              :checksum "E773BC66CF093065773D1120E5B4777A"
               :unzip true)
     (sift :move { #"^victory-.*/dist/victory\.js$"      "cljsjs/victory/development/victory.inc.js"
                   #"^victory-.*/dist/victory\.min\.js$" "cljsjs/victory/production/victory.min.inc.js" })
     (sift :include #{#"^cljsjs"})
-    (deps-cljs :name "cljsjs.victory")
+    (deps-cljs :name "cljsjs.victory"
+               :requires ["cljsjs.react"])
     (pom)
     (jar)))
--- a/victory/build.boot
+++ b/victory/build.boot
@@ -1,6 +1,7 @@
 (set-env!
   :resource-paths #{"resources"}
-  :dependencies '[[cljsjs/boot-cljsjs "0.5.1" :scope "test"]])
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.1" :scope "test"]
+                  [cljsjs/react "15.1.0-0"]])
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
--- a/victory/resources/cljsjs/common/victory.ext.js
+++ b/victory/resources/cljsjs/common/victory.ext.js
@@ -4,8 +4,12 @@ var Victory = {
   "VictoryAxis": function() {},
   "VictoryBar": function() {},
   "VictoryChart": function() {},
+  "VictoryGroup": function() {},
   "VictoryLabel": function() {},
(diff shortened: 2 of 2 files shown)


Deraen:
Yep, without require in deps.cljs the order in which the foreign libs are evaluated could be wrong.

Did you check if the extern file needs updates?
Changelog mentions at least one new component: `VictorySharedEvents`

euccastro:
@Deraen no, I didn't check, sorry.  I'm new to this area and I was just trying to get this to work; upgrading the victory.js version is just a remnant of things I tried (and didn't help).  I left the version upgraded because I thought it couldn't hurt.  I'll take a look at the advertised components and add any that are not there.

euccastro:
@Deraen: a safer alternative might be to leave the version alone and only update the deps.cljs.  If you'd be more comfortable merging that I'm happy to make that change.

Deraen:
I'm okay with merging this or the older version. Looks like the new version shouldn't have huge API changes anyway.

If you have time, the best thing would probably be to use a generator to upgrade the extern file. (There is a small chance that generator doesn't work for this library). I'm not sure how the current extern has been generated or if it's has been manually written. If you generate new extern file, this is also a good time to document how it has been created, like here: https://github.com/cljsjs/packages/blob/master/react-grid-layout/resources/cljsjs/react-grid-layout/common/react-grid-layout.ext.js

If you don't have time to check extern stuff, comment and I'll merge this without extern changes.

https://github.com/cljsjs/packages/wiki/Creating-Externs
https://github.com/cljsjs/packages/wiki/Updating-packages

euccastro:
Please don't merge just yet.  I've accidentally pushed the new externs before testing them.  I'll also add a comment on where I got them from.

This is the first I hear about extern generators.  I'll give that a try too, but I can't devote a lot of time to that.

euccastro:
Also, it seems this version of victory.js depends on react version 15.  Any way I can express this specifically?  I'd rather not bundle it with victory.js, since this is more of a change than I'm comfortable with ATM.

Deraen:
Correct way to depend on React (and specific version) is to add dependency to React: https://github.com/cljsjs/packages/blob/master/react-bootstrap/build.boot#L5

euccastro:
@Deraen I've added a dependency on the current react version in cljsjs as of this writing.  I've also tested that I can get the end result to work with reagent (after excluding the react version that comes with that), and that I can access all names I've added to the externs.  I've added a comment to the file as for where I got the list from.

I need to go, so I won't get to try automated externs generation today.

Deraen:
Thanks!

euccastro:
Thanks to you for the review and advice!

euccastro:
@Deraen unfortunately, I can't get VictoryLine to work for any version later than 0.6.1, although other components seem to work fine.  I think this library is very crippled if you can't make line graphs.  Just changing the version to 0.6.1 in the build.boot seems to fix the problem.  I didn't even have to remove from the externs the names that were only introduced in later versions, although I imagine that's a good thing to do anyway.

I'm not sure what one is supposed to do in this case.  I've dropped a note in the gitter channel of the victory.js project explaining the problem.  I'd like to roll back the recommended version of this library to 0.6.1 until this issue is fixed, because I think VictoryLine is more useful than whatever was added since 0.6.1.  Is that possible, once 0.8.0 has been published?  If not, can I at least publish a 0.6.1 with fixed dependencies?

Deraen:
Did you check if you are really using React 15?

Other dependencies (Reagent etc.) can overwrite React version. Check `lein deps :tree` or `boot show -d`.

Dropping the version to 0.6.1 is also fine.

euccastro:
Yes, I isolated the issue with the following minimal .html which doesn't involve Clojurescript or JSX, where I substituted the various versions of `victory.js` right out from their release .zip files.  The page below renders fine up to 0.6.1, but breaks with any later release.

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <title>VictoryLine test</title>
  </head>
  <body>
    <div id="react-app"></div>
    <script src="https://fb.me/react-15.1.0.js"></script>
    <script src="https://fb.me/react-dom-15.1.0.js"></script>
    <script src="victory.js"></script>
  </body>
  <script>
   var rootElement = React.createElement('div', {}, React.createElement(Victory.VictoryLine, {}));
   ReactDOM.render(rootElement, document.getElementById('react-app'));
  </script>
</html>
```

Upstream has confirmed that they could reproduce the problem and that a fix is in the works.  0.6.1 seems obviously the most useful version to provide until this is fixed.  I'll make a PR now to drop the version, and a new one to upgrade when the issue is fixed.

PR #610, opened 2016-06-12
[ADDED] Skel

flyboarder:
Added skel 3.0.1
http://skel.io

*missing externs


flyboarder:
@Deraen this now has externs 👍

PR #614, opened 2016-06-13
[auth0-cljs] Move to version 9.2.1

no-man-is-an-island:


Deraen:
Did you check if the externs need to be updated?

no-man-is-an-island:
Yeah -  it all seems to be the same. Think the major version release is to enable their new password reset mechanism.

vallard:
Is there a v10 now?  I see it on the API pages.  Would love to see more examples of how to use this as I'm new to cljs and having issues getting it working properly.  Looks like @no-man-is-an-island is making a library for this here: https://github.com/entrepreneur-first/auth0-cljs  I've tried to follow that too.  Thanks for all the work on this!

PR #615, opened 2016-06-14
RRule.js prerelease

eyelidlessness:
Update:

Includes support for `RRuleSet` and `rrulestr`.

### DIFF
--- a/rrule/build.boot
+++ b/rrule/build.boot
@@ -4,9 +4,9 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +commit+ "838226d")
-(def +lib-version+ (str "2.1.0-" +commit+))
+(def +lib-version+ "2.1.0-838226d")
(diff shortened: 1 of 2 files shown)


rrule/build.boot
```diff
@@ -4,7 +4,8 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "2.1.0")
+(def +commit+ "838226d")
+(def +lib-version+ (str "2.1.0-" +commit+))
```
Deraen: The build automatic is harded to read only +lib-version+ and +version+ using just regex so it can't evaluate this.

This should work:
```clj
(def +lib-version+ "2.1.0-838226d")
(def +version+ (str +lib-version+ "-0"))
(def +commit+ (second (re-matches #".*-(.*)" +lib-version+)))
```

eyelidlessness: Updated!

PR #616, opened 2016-06-18
upgrade victory.js to v0.9.0

euccastro:
This version fixes the problem with VictoryLine which had us pinning this library to v0.6.1.

Update:


victory/resources/cljsjs/common/victory.ext.js
```diff
@@ -1,24 +1,18 @@
...
-// This list should be good for a 0.7.0 release, should you want to make it. For
-// 0.8.0 or later, the VictorySharedEvents entry should be uncommented at least.
 
 var Victory = {
   "VictoryAnimation": function() {},
   "VictoryArea": function() {},
   "VictoryAxis": function() {},
   "VictoryBar": function() {},
   "VictoryChart": function() {},
+  "VictoryContainer": function() {},
```
euccastro: The only new addition in v0.9.0.

PR #619, opened 2016-06-21
aws-sdk-js: Fix production aws-sdk.min.js path

dbrenden:
The production path for aws-sdk-js.min.js has a hyphen where there should be a dot, and as a result there is no production folder in the release.


Deraen:
Please update build number to get new release:

```
```

PR #624, opened 2016-06-27
[WIP] cljsjs/react-toolbox

lwhorton:
New package:

**Extern:** Written by hand.

### DIFF
--- a/react-toolbox/README.md
+++ b/react-toolbox/README.md
@@ -1,0 +1,28 @@
+# cljsjs/react-toolbox
+
+```clojure
+[cljsjs/react "1.0.1-1"] ;; latest release
+```
+
+[React Toolbox](http://react-toolbox.com/)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns my-ap.core
+  (:require [cljsjs.react-toolbox]))
+
+(def react-toolbox js/ReactToolbox)
+```
+
+Note that react-toolbox is designed to work hand-in-hand with webpack and Sass.
+This is quite cumbersome to setup with cljs, so cljsljs/react-toolbox publishes
+each react-toolbox component in the bundled form (styles included). To override
+component styles, take a look at [react-css-themr](https://github.com/javivelasco/react-css-themr).
+
+React, ReactDOM, and ReactAddonsCssTransitionGroup are the dependencies for
+react-toolbox. This cljsjs/react-toolbox supports loading these libraries via
+commonjs, commonjs2, or plain old `window.{dependency}`. Take a look at the webpack
+config inside `/resources` for more details.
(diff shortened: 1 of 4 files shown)


ducky427:
I think there are a few things wrong with package:

1. This package should have `cljsjs.react` or `cljsjs.react-with-addons` as a dependency.
1. There is no need for `deps.cljs` file.
1. Are you sure about the syntax of `react-addons-css-transition-group` as an external dependency in the webpack file?

lwhorton:
1. Thanks. I wasn't sure if that should or should not be a dep.
2. I'm unclear on why some packages have a deps and some do not (and why does this one not?).
3. No, you're correct. I thought css-transition-group was an external package but after poking around on the react docs (not very clear) and confirming in the code, it seems it's still under `React.addons`... and thus the webpack should not expose that as an external dep.

lwhorton:
Per the help of @ducky427 and @martinklepsch on clojarians.slack, I think the latest rebase has fixed all the issues.

lwhorton:
More issues fixed:

1. failure to `:requires` react has been fixed. **note** that I'm not 100% sure if this should be `cljsjs.react-with-addons` (or some variant `react.with-addons`, etc), but addons seems to be packaged under `cljsjs.react`. I could find no example in other packages mentioning -with-addons.
2. webpack build must list the `css-transition-group` as an external, but this is awkward because `react-with-addons` brings the dependency, and `-css-transition-group` has to "read it in" from the environment.
3. removed deps.cljs in the sift task, which is not needed

ducky427:
LGTM. 👍

lwhorton:
Actually using this in an implementation I ran into an issue with selector specificity. Since in cljsjs land we cannot easily use webpack to write sass to customize components, this library pre-builds components a single importable object. Those pre-built components, when imported, attach a `<style>` tag to the head, thus achieving "plug-n-play" interop. However, by default the style tags are added to the bottom of the head... which causes havoc when providing custom themes from some css module elsewhere. This simple fix puts the tags at the head, so user-land css-modules can effectively overwrite predefined styles.

Deraen:
I don't think it is a good idea to automatically add stylesheet to HTML. Instead just provide single css (and less/sass if possible) file in the package, that users can include on their own, using [ring-cljsjs](https://github.com/Deraen/ring-cljsjs) or boot-sass/less etc.

lwhorton:
I absolutely agree, and thats part of why this remains a WIP. The problem unfortunately is a little more effort to solve than I hoped. The lib uses sass preprocessing with webpack as almost a prerequisite. [Check here](https://github.com/react-toolbox/react-toolbox/issues/597#issuecomment-231174695) for some more thoughts. I agree with the  sane solution listed (although a lot of work): expose a stylesheet using bem instead of css modules, then allow consumers to override as needed. 

~luke

lwhorton:
I no longer believe this is a good idea. css modules intermixed with cljs, webpack preprocessors, and inline style tags is just too complex and doesn't feel very clojure-ey. I can't come up with a good scheme to export components appropriately so they're flexible without imposing a *whole* lot of implementation burden details on to consumers. Re-creating the entire library inside my app as straight hiccup took about 4 days of effort, for those curious, and the outcome was much cleaner.

PR #630, opened 2016-07-01
[d3] Updated to v4.0.0

ducky427:
Update:


ducky427:
The API seems to have changed considerably. I've regenerated the externs.

PR #638, opened 2016-07-11
[typeahead-bundle] Add dependency on jquery

flosell:
Update:

**Dependencies:** added dependency on jQuery.

Typeahead depends on jQuery. If the dependency declaration is missing in the package, typeahead might break because it gets loaded before jQuery.


Deraen:
This should presumably also depend on the cljsjs/jquery package. Deps.cljs require alone doesn't add the maven dependency.

flosell:
fixed, does this look better?

Deraen:
Yep!

PR #645, opened 2016-07-19
Update params in react-dom

dannycochran:
Update:

Caught these misnamed parameters when compiling.


ducky427:
LGTM!

PR #650, opened 2016-07-22
Download externs from npm in boot.build packaging -- 3.2.0

timothypratley:
Update: 3.2.0

Pulls externs from NPM as part of the build.boot packaging

image


firebase/build.boot
```diff
@@ -1,11 +1,10 @@
 (set-env!
-  :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.2" :scope "test"]])
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "3.0.5")
-(def +version+ (str +lib-version+ "-rc2-0"))
+(def +lib-version+ "3.2.0")
+(def +version+ +lib-version+)
```
crisptrutski: Let's keep a `-0` suffix on lib-version to follow CLJSJS conventions

firebase/build.boot
```diff
@@ -17,11 +16,13 @@
...
-    (download :url (str "https://github.com/firebase/firebase-bower/archive/v" +lib-version+ ".zip")
-              :checksum "559003EA8EF0375A16140BBB8AB87905"
-              :unzip true)
-    (sift :move {#"firebase-bower-([\d\.]*)/firebase.js" "cljsjs/development/firebase.inc.js"})
-    (sift :include #{#"^cljsjs"})
-    (deps-cljs :name "cljsjs.firebase")
-    (pom)
-    (jar)))
+   (download :url (str "http://registry.npmjs.org/firebase/-/firebase-" +lib-version+ ".tgz")
+             ;;:checksum "29ab75208d45f96e312121d1f546fceec77653de"
```
crisptrutski: These checksums should still to be stable, so don't see a reason to comment out

crisptrutski:
This certainly eases future maintenance, thanks!

crisptrutski:
Curious if the same update is possible for `firebase-node`, or even better if they're publishing cross platform  releases now that it's on npm?

firebase/build.boot
timothypratley: I agree, but it fails...
Is a shasum different from a checksum?
```
dist: {
shasum: "29ab75208d45f96e312121d1f546fceec77653de",
tarball: "https://registry.npmjs.org/firebase/-/firebase-3.2.0.tgz"
},
```
I'm thinking of removing the comment to avoid confusion

firebase/build.boot
timothypratley: good catch! done

timothypratley:
firebase-node.js is inside that same firebase npm repository but I don't see any externs for it specifically... so I suspect they haven't released any yet (or could the same externs apply to both)? I haven't used firebase-node (yet)... so not sure :)

martinklepsch:
Wow, so cool their packaging externs as well now.

flyboarder:
@timothypratley the externs provided are for both versions of firebase last i checked. Also watch the rc version of the package, latest is `3.2.0-rc.2`.

timothypratley:
@flyboarder Oh that's great! It might be sensible then to include `firebase-node` in this package and deprecate the firebase-node folder, as it would be closer to how the npm is packaged?

flyboarder:
@timothypratley But we need to keep loading separate between the client and node packages. ie. maintain cljsjs.firebase and cljsjs.firebase-node

timothypratley:
What I mean is that the Closure Compiler will only compile in the client or the node code if the user requires it from there code, so there is no need to package them separately... so it might be slightly more convenient to do it that way - just an idea :)

timothypratley:
After looking into this further, what I said is wrong. The problem being that deps.cljs defines `:externs` globally for the package... so even though you can require just firebase or firebase-node, you'll always get the externs for both. Which is not quite right. It's mostly fine because the externs overlap, but I'm going to open a pull request to split the externs into the packages that they belong in. I don't think this affects compilation, but it seems like it would be best to only have the externs that apply to the package you want.

timothypratley:
So after looking into this further, it seems like it is possible except that the `deps-cljs` task currently will only take a single 'name' for what will be provided. This means we'd have to not use `deps-cljs` and provide a deps.cljs file manually edited instead if we wanted to go this route... or update deps-cljs to allow overriding the `:provides` (currently it is taken from name only).

As a side note I got a bit confused about externs because there is a 'server-auth' which I incorrectly took to be node specific. I don't really get why they are separate files, but it seems fine. Cursive gets confused by it so I'll raise an issue with Colin about that.

Deraen:
It is okay to provide either manually written deps.cljs or write additional tasks that modify the file: https://github.com/cljsjs/packages/blob/master/highlight/build.boot#L25-L41

timothypratley:
Ah cool thanks!
I've got pull request going here that uses a manual deps.cljs:
https://github.com/cljsjs/packages/pull/656
and will look into whether I can make use of that snippet you linked instead.

PR #656, opened 2016-07-28
Adds firebase-node and bumps to 3.2.1

timothypratley:
Pending discussion:

This would bring firebase-node into the firebase package (if this is held to be desirable)...
This is closer to how npm packages firebase+firebase-node.

Notably I am not a firebase-node user, so please complain if this seems bad.
(I have not adequately test firebase-node, I need to spend some time on it... soon I hope! If you do use firebase-node and can try it out that would be much appreciated)
cljsjs/firebase itself seems fine for my usage of it.


timothypratley:
I'm thinking of concatenating all the externs files into a single .ext.js file to avoid https://github.com/cursive-ide/cursive/issues/1477

Deraen:
@timothypratley What is the status of this PR? Should I merge this?

timothypratley:
@Deraen I think it can move ahead, but also it wouldn't hurt to wait a couple of days for comments or objections.

flyboarder:
lgtm 👍

PR #662, opened 2016-07-29
[highlight] Add instructions about adding theme CSS to projects

anmonteiro:
Add instructions on how to add CSS to highlight.js users

### DIFF
--- a/highlight/README.md
+++ b/highlight/README.md
@@ -15,6 +15,10 @@ you can require the packaged library like so:
   (:require cljsjs.highlight))
 ```
 
+Refer to [this](https://github.com/cljsjs/packages/wiki/Non-JS-Assets) wiki page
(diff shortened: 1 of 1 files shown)


Deraen:
I think it would be better to link to this wiki page: https://github.com/cljsjs/packages/wiki/Non-JS-Assets

Additionally you could mention this pure `sift` solution before `from-jars` task example.

anmonteiro:
I was just following the lead of the CodeMirror's [README](https://github.com/cljsjs/packages/blob/master/codemirror/README.md), which has similar instructions. I'm happy to change it to link to the wiki, if that's how I should proceed.

Deraen:
Yeah, it is best to link to the wiki page from every readme.

anmonteiro:
@Deraen updated the PR with your feedback

PR #663, opened 2016-07-30
[react] Updated to v15.3.0

ducky427:
Update:

I am not a jsdoc expert. So need to check if the extern for [`React.PureComponent`](https://github.com/facebook/react/blob/master/src/isomorphic/modern/class/ReactPureComponent.js) is accurate.

cc @Deraen, @martinklepsch

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -1,7 +1,7 @@
 # cljsjs/react
 
 ```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
Seems fine. CloureScript doesn't even use the type information.

PR #669, opened 2016-08-01
[react-input-autosize] Update to v1.1.0

svdm:
Update:


Deraen:
Btw. just tried to use this, but looks like new react-input-autosize (>1.0) doesn't work without `require`: 

```
react-input-autosize.inc.js:1 Uncaught Error: Cannot find module 'react'
    at s (react-input-autosize.inc.js:1)
    at react-input-autosize.inc.js:1
    at Object.2../factory (react-input-autosize.inc.js:742)
    at s (react-input-autosize.inc.js:1)
    at react-input-autosize.inc.js:1
    at Object.<anonymous> (react-input-autosize.inc.js:1875)
    at Object.14.create-react-class (react-input-autosize.inc.js:2035)
    at s (react-input-autosize.inc.js:1)
    at e (react-input-autosize.inc.js:1)
    at react-input-autosize.inc.js:1
    at react-input-autosize.inc.js:1
```

> You can also use the standalone build by including dist/AutosizeInput.js in your page. If you use this, make sure you have already included React and Underscore. (they must be available through a browserify-style require() call, standalone support is coming soon)

Deraen:
Oops, sorry, 1.1.0, works, 1.1.2 is broken (1.1.1 is not found on npm...).

svdm:
Odd, the same blurb about require was also in the 1.1.0 readme. It looks like there wasn't actually a require call in the build at that point though: https://github.com/JedWatson/react-input-autosize/blob/v1.1.0/dist/react-input-autosize.js

It looks like most of the `require`-ing in newer versions is done in the prop-types and create-react-class deps that are packed into the standalone build. The actual lib only has requires for those two: https://github.com/JedWatson/react-input-autosize/blob/v1.1.3/dist/react-input-autosize.js#L1873

Perhaps there's some gulp/browserify/whatever magic that will compile those two require calls to end up like the `react` require (ie. just using the global `React`). If someone can point me in the right direction I'm happy to put in some work on the build process for this package. I can draw inspiration from other cljsjs packages that do their own builds but I'd prefer not to go in blind.

Deraen:
If the package provides standalone files, it is best to use those, I opened issue at react-input-autosize so hopefully they can fix the build: https://github.com/JedWatson/react-input-autosize/issues/95

PR #680, opened 2016-08-05
[react-datepicker] Update to version v0.28.2

adstage-david:
Note that the build process has grown more complicated:

* it requires an npm install/grunt build
* it requires a gem install to build the scss files.

Also, a missing dependency on react-onclickoutside has been added as a
separate cljsjs package that this package now requires:

https://github.com/cljsjs/packages/pull/679

Update:

**Extern:** I updated the extern with generator


adstage-david:
(Note the build appears to be failing failing due to the requirement on #679)

react-datepicker/build.boot
```diff
@@ -1,13 +1,18 @@
 (set-env!
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]
-		  [cljsjs/react "0.14.3-0"]
-                  [cljsjs/moment "2.10.6-0"]
+                  [cljsjs/react "15.3.0-0"]
+                  [cljsjs/moment "2.10.6-4"]
+                  [cljsjs/react-onclickoutside "4.9.0"]
```
Deraen: This dependency is missing the build identifier: `-0`

adstage-david:
Whoops - thanks for catching that! Fixed and pushed up, should build clean now.

PR #688, opened 2016-08-08
[react-burger-menu] update to v1.9.9

worrel:
Update:

Not bumping all the way to latest v1.10.2 since that now depends on snapsvg-cjs which is a re-packaging of snapsvg & I think that violates the "no package against forks".

The radium dependency is now required at runtime.


worrel:
Any love for this? It's a small change to the packaging, but results in a big improvement (the ability to customize the burger menu icon in particular, which defaults to quite a large one)

PR #689, opened 2016-08-08
[react-onclickoutside] Add missing dependency

adstage-david:
I referenced `cljsjs.object-assign-shim` in the deps-cljs, but did not include it in the actual dependencies at the top of the file. This should fix the bug reported in slack with the react-datepicker version 0.28.2 where someone had to manually include `[cljsjs/object-assign-shim "0.1.0-0"]` to use that version

Update:


Deraen:
If you want updated package, please update build identifier:

```clj
 ```

adstage-david:
Whoops! Amended.

PR #694, opened 2016-08-12
Add fuse

adolby:


adolby:
I think this failed because CircleCI couldn't connect to clojars.

PR #702, opened 2016-08-21
[recharts] Add Recharts library

anmonteiro:
New package:

**Extern:** Generated.

### DIFF
--- a/recharts/README.md
+++ b/recharts/README.md
@@ -1,0 +1,18 @@
+# cljsjs/recharts
+
+[](dependency)
+```clojure
+[cljsjs/recharts "0.13.3-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the Clojurescript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.recharts))
+```
+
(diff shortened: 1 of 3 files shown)


ducky427:
My comments:

1) `README.md` has a reference to `cljsjs/bootstrap`. It should probably be `cljsjs/recharts`.
2) This library seems to have no dependency to react. Is react packaged with it or is that an oversight?

anmonteiro:
1) fixed

2) added `:requires [cljsjs/react]`, definitely an oversight

ducky427:
@anmonteiro, really sorry but 2) isn't done. You need to add it in: `build.boot` in `set-env!` `dependencies` as well, much as [this](https://github.com/cljsjs/packages/blob/master/react-virtualized/build.boot#L4)

anmonteiro:
@ducky427 gotcha, wasn't aware. Which version of React should it depend on? Does it depend on the library itself, or should it be the latest, or something else?

ducky427:
I'd say the version the library depends on. 

React API is pretty stable. So shouldn't be an issue. but still best to stick with what `rechart` authors have tested against.

anmonteiro:
@ducky427 updated the PR with React 15.3.1-0
Recharts specifies peer dependencies of "^14.0.0 || ^15.0.0" so the latest is OK

ducky427:
Thanks!

PR #705, opened 2016-08-23
[stats] Add stats.js - a library for JavaScript performance monitor

nvbn:
New package:
**Extern:** Written by hand.

### DIFF
--- a/stats/README.md
+++ b/stats/README.md
@@ -1,0 +1,18 @@
+# cljsjs/stats
+
+[](dependency)
+```clojure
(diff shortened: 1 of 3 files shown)


ducky427:
The externs for this package are missing quite a few functions and it is likely to confuse the end user if compiled with advanced optimisations.

nvbn:
Oops, forgot to check extern. Now it works with advanced optimisations.

ducky427:
cheers!

PR #709, opened 2016-08-25
Adding react-sticky to cljsjs

hlolli:
New package:

**Extern:** Generated.


hlolli:
There's a bug in this js library, they are using deprecated findDOMnode that can be easily fixed with replacing: `var node = _reactDom2.default.findDOMNode(this);` to  `var node = ReactDOM.findDOMNode(this);`  so I make another pull request soon.

Deraen:
I don't understand the problem:

```
var node = _reactDom2.default.findDOMNode(this);
var node = ReactDOM.findDOMNode(this);
```

are the same thing. First one just accesses ReactDOM through Webpack module wrapper.

PR #712, opened 2016-08-30
[Auth0] Update to 10.2.1-0

AdamFrey:
Update:


agzam:
I think this is breaking when optimizations enabled. :(

PR #723, opened 2016-09-04
[flatpickr] Add flatpickr

simonrobson:
**Extern:** Generated. With manual additions


Deraen:
Thanks. Good notes on the extern file!

PR #724, opened 2016-09-05
[material] upgrade to mdl 1.2.0

sonwh98:
upgrade to mdl 1.2.0

### DIFF
--- a/material/build.boot
+++ b/material/build.boot
@@ -4,8 +4,8 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "1.1.3")
-(def +version+ (str +lib-version+ "-1"))
+(def +lib-version+ "1.2.0")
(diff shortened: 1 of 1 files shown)


martinklepsch:
Hey @sonwh98 👋  Have you checked if the externs need updating? The [release page](https://github.com/google/material-design-lite/releases) mentions a new type of component but I'm not sure if that necessarily means that the API has changed as well.

sonwh98:
Hi @martinklepsch I tested locally using the new components  without any problems

Deraen:
Did you test with `:optimizations :advanced`?

flyboarder:
@Deraen @sonwh98 I checked the commit which introduced the new component and no javascript was included for it to function.

sonwh98:
yes I have tried with advance optimization. However as @flyboarder  says no javascript needed for the new component

PR #726, opened 2016-09-05
[d3] Bump to the latest patch release and fix some missing externs

jszakmeister:

I did my best to expose much of the bits from scale and more bits from axis.  I suspect there is much more missing from the D3 externs. :-(


jszakmeister:
Thanks @Deraen!

PR #730, opened 2016-09-07
[react-flexbox-grid] 0.10.2-0: initial release

madvas:
New package:

**Extern:** Written by hand.


Deraen:
Looks like this build currently includes LOADS of unncessary files, the resulting jar is 40 megabytes and looks like it is rejected by Clojars.

Deraen:
Should be fixed by https://github.com/cljsjs/packages/commit/5cc54fa2cb6272660f818a104dacae24abefeb5e

PR #744, opened 2016-09-16
[highstock] 4.2.2->4.2.6

lwhorton:
Update:

**build.boot:**
Changed `:download` task from `http` to `https`.

**Extern:** 
One API item was added at `4.2.4` [changelog](http://www.highcharts.com/documentation/changelog#highstock)
I updated the extern by hand.

**Comments:**
Should we also update highcharts in unison? I took a look at the [changelog](http://www.highcharts.com/documentation/changelog) and there's quite a bit of additional configurations added, but those configs don't need to be listed in the externs file. 

One piece I wasn't quite sure about (relating to high*charts* not high*stock* is the `chart.events`. The existing externs file doesn't list anything related, so I'm not sure if it's not needed or it was missing the whole time. Perhaps @jimlar can comment?

### DIFF
--- a/highstock/build.boot
+++ b/highstock/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "4.2.2")
+(def +lib-version+ "4.2.6")
 (def +version+ (str +lib-version+ "-0"))
 
 (task-options!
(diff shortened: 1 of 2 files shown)


lwhorton:
Another thing I forgot to mention, and again maybe @jimlar can comment -- since `4.2` highstock/charts doesn't require jquery anymore. Can we scrap that `:requires` in the build.boot?

Deraen:
If it works for you, go ahead. Previous version was invalid anyway, as it requires cljsjs.jquery namespace but doesn't depend on cljsjs/jquery package.

lwhorton:
👌

PR #760, opened 2016-09-27
[immutable] Add immutable package

kommen:
New package: immutable https://github.com/facebook/immutable-js

**Extern:** Generated.

This dependency will be needed in a later pull request which adds a package for http://github.com/ianstormtaylor/slate

Question: Do you prefer `immutable` or `immutable-js` as the package name?


miikka:
`immutable` seems fine and is in accordance with [what the wiki says](https://github.com/cljsjs/packages/wiki/Creating-Packages#3-create-a-directory-structure-for-the-package).

PR #778, opened 2016-10-12
[quill] update to v1.0.6

smogg:


Deraen:
Have you tested this with advanced compilation?

The extern file seems to miss all methods of Quill API: http://quilljs.com/docs/api/

smogg:
I created a new project and run an uberjar with the following setup for cljsbuild:

```clojure
{:source-paths ["src/cljs" "src/cljc" "env/prod/cljs"]
             :compiler
             {:output-to "target/cljsbuild/public/js/app.js"
              :output-dir "target/uberjar"
              :optimizations :advanced
              :pretty-print  false}}
```

I do seem to have access to api methods, for example I was able to use `deleteText` and it seems to be in the compiled js:

```
...
,value:function(){this.selection.setRange(null)}},{key:"deleteText",value:function(t,e,n){var r=l(t,e,n),o=u(r,4);t=o[0],e=o[1],n=o[3];var i=this.getSelection(),s=this.editor.deleteText(t,e,n);return 
```

Is there some other way I should test that?

Deraen:
What is the cljs source for deleteText call?
Are you using other cljsjs libs?

smogg:
I am calling Quill and testing its API from the browser console.
The project is a fresh `lein new reagent ...` template and I pushed it here: https://github.com/smogg/quill-test

smogg:
So, I'm not an expert when it comes to how things are compiled, but maybe the reason for why api methods are still in compiled js is because of the way Quill source code was written, for example `deleteText`:

```
...
key: 'deleteText',
value: function deleteText(index, length, source) {
...
```

Deraen:
Ah right, so the source code of library is never changed by Closure. Extern affect to code generated by ClojureScript compiler. E.g. when you call `(.deleteText quill ...)` from Cljs code. This call will be manged by Closure to something like `a.b.call(null, ...)` (where a = quill and b = deleteText), which will break because in foreign the method is still deleteText.

smogg:
Roger that, thanks for explanation. I used a different generator and it looks like it included the API methods

Deraen:
Thanks, these externs look better.

smogg:
There might have been a mistake on my end and for some reason the extern generated was actually corrupt (hence missing a bunch of stuff). Since Quill already updated to 1.1.0 I submitted another PR fixing this

PR #785, opened 2016-10-16
[material-ui] Fix missing extern for autocomplete filter options

edannenberg:
Update:

Using filter options for the autocomplete field caused render errors with enabled advanced compilation.


Deraen:
Ping @madvas.

I think material-ui extern is missing lots of stuff? Are you using some alternative way to go around this?

madvas:
This is ok to merge. Yeah, material-ui might be missing some externs, as it's not anyhow easy to obtain or generate from docs.

PR #787, opened 2016-10-17
[quill] Update to 1.1.0

smogg:
Update:


Deraen:
This extern doesn't contain e.g. `deleteText` method at all.

quill/resources/cljsjs/quill/common/quill.ext.js
```diff
@@ -49,52 +25,1859 @@ var Quill = {
...
+  "canAdd": function () {}
+};
+Quill.imports.parchment.Attributor.Store.prototype = {
+  "attribute": function () {},
+  "build": function () {},
+  "copy": function () {},
+  "move": function () {},
+  "values": function () {}
+};
+Quill.imports.blots / block.prototype = {
```
Deraen: This doesn't look like valid JS.

smogg:
I updated the externs manually and tested all the api methods from cljs. Thanks for your patience @Deraen

PR #788, opened 2016-10-18
[quill] Externs were not correct for Quill themes

smogg:
This fixes a problem where Quill would complain about its themes not being registered


Deraen:
To get a new release, you should also bump build identifier (version number) on build.boot.

smogg:
Done, thanks

PR #797, opened 2016-10-26
[filesaverjs] Update to 1.3.3

salax:
Update:


filesaverjs/build.boot
```diff
@@ -20,7 +20,7 @@
 (deftask package []
   (comp
    (download :url (str "https://github.com/eligrey/FileSaver.js/archive/" +lib-checkout+ ".zip")
```
Deraen: There is tagged release (https://github.com/eligrey/FileSaver.js/archive/1.3.3.zip) available using version 1.3.3 so please use that instead of sha.

PR #798, opened 2016-10-26
[accounting] Updates externs & bumps version number

tuddman:
Update:

minified build now compiles cleanly


Deraen:
Please fix the conflict in README.

PR #802, opened 2016-11-01
[moment.js] Upgrade 2.10.6 -> 2.15.2

Gonzih:
I did not see any big api changes, only one new method was added. I tested similar upgrade recently on my own projects, but I use small subset of moment, so can't say for sure if api changed.


Gonzih:
Thx! <3

PR #803, opened 2016-11-01
Added externs for d3.radialLine

achesnais:
Update:
**Extern:** I updated the D3 extern by hand.


Deraen:
To get a new release, you should also update build identifier in the version number. As I already merged this before I noticed this, you create separate PR for this.

Deraen:
Or nevermind, I merged another d3 PR which changed version number.

PR #805, opened 2016-11-01
[moment.js] Add missing externs

guilherme-teodoro:
Update:

**Extern:** Add `defineLocale`


Deraen:
This conflicts with another recent PR (perhaps one made latter than this, sorry).

PR #825, opened 2016-11-14
[victory] Update to 0.13.7

mthomure:
Build from NPM package, since dist files are no longer available in archive.


mthomure:
Hi all,

Something's wrong, here. The build works fine for me locally, but seems to
hit an error when running on CircleCI. The relevant output is here:

https://circleci.com/gh/cljsjs/packages/1265#config/containers/0

I've almost certainly dorked something, and would appreciate any pointers
to get this build fixed.

Thanks,
Mick

P.s. Thanks for CLJSJS!

Deraen:
@mthomure The problem was capital `V` in `require` call. Npm modules usually (always?) start with lower case letter.

While searching for this, I did a few changes to  make build faster and to fix some warnings: https://github.com/cljsjs/packages/commit/3ba4dfd9fd1cd727488f057a528394f8b50a3fc2

mthomure:
Ugh, sorry for the dumb mistake. Very new to the Javascript world. Thanks for the fix!

PR #827, opened 2016-11-14
Add libtess package

skrat:
New package:

**Extern:** Not needed as this is Closure library.

### DIFF
--- a/libtess/README.md
+++ b/libtess/README.md
@@ -1,0 +1,7 @@
+# cljsjs/libtess
+
+[](dependency)
+```clojure
+[cljsjs/libtess "1.2.2-0"] ;; latest release
+```
+[](/dependency)
--- a/libtess/build.boot
+++ b/libtess/build.boot
@@ -1,0 +1,36 @@
+(set-env!
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +lib-version+ "1.2.2")
+(def +version+ (str +lib-version+ "-0"))
+
+(task-options!
+ pom {:project 'cljsjs/libtess
+      :version +version+
+      :description "Polygon tesselation library"
+      :url "https://github.com/brendankenny/libtess.js"
+      :scm {:url "https://github.com/brendankenny/libtess.js"}
+      :license {"SGI" "https://raw.githubusercontent.com/brendankenny/libtess.js/gh-pages/LICENSE"}})
+
+(def github
+  "https://raw.githubusercontent.com/")
+
+(deftask package []
+  (comp
+   (download
+    :url (str github "brendankenny/libtess.js/" +lib-version+ "/libtess.cat.js")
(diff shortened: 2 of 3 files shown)


skrat:
Can anyone assist me on this one? This is a Closure library, I'm not entirely sure how to package it as extracting externs seems redundant to me.

Deraen:
Concatenated or optimized file is no longer Closure library, you would need include the original source files.

And while the project uses Closure compiler to optimize the output, I don't see anything in source files to indicate that this would be a Closure library, files are Closure compatible but don't seem to provide any Closure modules (no `goog.provide` calls).

skrat:
@Deraen Thanks for looking into it. So my idea was using the concatenated file twice: a) as a non-minified development version b) as externs file

However I didn't find a way to make boot copy the source file to 2 locations. What do you think is best way to proceed?

Deraen:
Writing a proper extern file. Using code file as extern file is not good practice, no matter if this is Closure compatible or not.

You can use https://github.com/brendankenny/libtess.js/blob/gh-pages/build/closure_exports.js as basis. Just replace all mentions on the functions and values with empty functions or objects, like this:

```js
this['libtess'] = {
  'GluTesselator': function() {},

  // enums
  'windingRule': {
    'GLU_TESS_WINDING_ODD': function() {},
    'GLU_TESS_WINDING_NONZERO': function() {},
...
```

skrat:
@Deraen Sounds good. I'm just wondering why is it not a good practice. IMHO duplicating the annotations is redundant and not a good practice and Closure compiler is completely fine with using externs containing body functions and values.

Deraen:
Because externs are global and any, even internal and private, name (function, variable name) means that name can't be changed anywhere in the code being optimized.

Not sure how bit this problem is, but for example this project contains private functions: https://github.com/brendankenny/libtess.js/blob/gh-pages/src/mesh.js#L491

Including that file as extern would mean that name `makeFace_` is never optimized, which would lead to lesser compression.

Btw. had another look at closure_exports.js and files, not sure how this library works because closure_exports.js doesn't seem to refer e.g. libtess.mess at all.

skrat:
I think closure_exports.js is used by Closure compiler to export public API, and as you said, `libtess.mesh` is private and therefore the author of the library didn't include it in the exports.

skrat:
@Deraen Thanks for explaining the source-as-externs problem, it does make sense now. I added the externs. Does it look mergeable now?

skrat:
@Deraen I don't think this CI failure

```
User limit of inotify watches reached
```

has anything to do with my PR.

Deraen:
Yep, that is just a random error.

Deraen:
I don't know enough about libtess to be able to know if the externs are complete, but I guess they are enough for now. Mostly I'm wondering about functions like https://github.com/brendankenny/libtess.js/blob/gh-pages/src/mesh.js#L47, which are not referred in closure_exports.js.

skrat:
@Deraen Thank you :heart: as of the mesh functions, I don't know about it too, but it doesn't seem to be part of the public API. Public API usage is demonstrated in the `examples/`.

Deraen:
Yeah, I checked examples but didn't notice anything using mesh functions directly. Anyway, just create PR if you notice you need to add some functions to externs.

PR #830, opened 2016-11-17
[victory] fix build error

mthomure:
Build for this package was broken. This PR downloads prebuilt JS from CDN, instead of using NPM package.


mthomure:
Ooos. @Deraen looks like you beat me to it. Thanks!

Deraen:
Using cdn version could be preferable to building our own versions. At least they are tested better than our custom webpack configuration.

Deraen:
So feel free to reopen this and update build version once again.

PR #832, opened 2016-11-17
[canvas2image] canvas2image library added into cljsjs packages

joekurta:
New package:

**Extern:** Generated.
**Extern:** Written by hand.

### DIFF
--- a/canvas2image/README.md
+++ b/canvas2image/README.md
@@ -1,0 +1,9 @@
+# cljsjs/html2canvas
+
+[](dependency)
+```clojure
+[cljsjs/canvas2image "0.1-0"] ;; latest release
+```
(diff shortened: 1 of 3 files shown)


Deraen:
I'm usually quite skeptic about libraries without proper releases (no git tags, no npm releases), but as long as the build is deterministic I can merge this.

Btw. did you know that you can just use `canvas.toDataURL("image/png")` to create data URI you can use with `img` elements to display the contents? Downloading the image a bit more complex so library can be useful for this.

Some alternatives also (neither of these has proper releases either...)
- https://github.com/gillyb/reimg
- https://github.com/kaxi1993/canvas-to-image (this can set the filename for download, which I don't think canvas2image supports)

PR #833, opened 2016-11-17
Fix exports for libtess (node.js compat)

skrat:
This fix makes it work when compiling to `nodejs` target.

### DIFF
--- a/libtess/build.boot
+++ b/libtess/build.boot
@@ -5,7 +5,7 @@
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
 (def +lib-version+ "1.2.2")
-(def +version+ (str +lib-version+ "-0"))
+(def +version+ (str +lib-version+ "-1"))
 
 (task-options!
  pom {:project 'cljsjs/libtess
@@ -19,7 +19,7 @@
   "https://raw.githubusercontent.com/")
 
 (def export
-  "(window||global)")
+  "(((typeof window != 'undefined') && window) || global)")
 
 (deftask package []
   (comp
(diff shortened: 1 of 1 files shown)


Deraen:
Huh, I missed this replace content business previously.

Can you explain these two regexes to me: `#"(.|\n){70}$"` `#"(.|\n){147}$"`

These look like something which can very easily break between updates?

Deraen:
Also, did you try the code without replacing `this.libtess`? In browser `this = window` unless it has been bound to something else, so I think the original file should work. Not sure about Node.

skrat:
@Deraen Those first regex are removing the CommonJS exports, because they cause havoc, when I compiled a CLJS project that is later on consumed by browserify. Yes I did try the code without replacing `this.libtess` assignment, and it doesn't work, since Closure compiler is wrapping the whole thing in a closure where `this` refers to `null`.

I could only get it working in all the environments (browser, node.js, advanced optimizations) with this patch.

Deraen:
Okay. But what those regex are supposed to match, looks to me like they would be trying to select line on certain length?

skrat:
They are removing from end of the input, since the exports are at the end. The input is checksummed so it matches exact number of characters from the end. You diff the input with the sifted files to see what it does.

Deraen:
Okay, so regexes will break if when version is changed.

Please add a huge warning to a line before or after version number about updating/checking regexes after version change.

skrat:
@Deraen I don't think it will necessarily break, as those exports are unlikely to change, since they export just one (`libtess`) object. Also, libtess.js is feature complete, there will only *ever* be patch level updates.

Anyway, thanks for your time and for looking into it, also thanks for the concerns, they're all perfectly valid.

PR #849, opened 2016-11-27
[New] Feathersjs

flyboarder:
Feathersjs packaged for cljsjs
Version: 1.8.0 (latest)
Externs: Generated


martinklepsch:
Feathers.js looks pretty cool! Kind of like Horizon from the Rethink folks? :)

flyboarder:
just like horizon but I think it will do better.

PR #860, opened 2016-12-01
Added blend4web externs/support files

mikebelanger:
New package:
**Extern:** Generated (from an existing (non-clojure) project)

### DIFF
--- a/blend4web/build.boot
+++ b/blend4web/build.boot
@@ -1,0 +1,216 @@
+(def +lib-version+ "16.11")
+(def +version+ (str +lib-version+ "-1"))
+
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all]
+         '[boot.task.built-in :refer :all])
+
+(task-options!
+ pom  {:project     'cljsjs/blend4web
+       :version     +version+
+       :description "Blend4Web -- Javascript WebGL Framework by Triump LLC"
+       :url         "http://www.blend4web.org/"
+       :license     {"GPL3" "https://github.com/TriumphLLC/Blend4Web/blob/master/license/GPL-license.txt"}
+       :scm {:url   "https://github.com/cljsjs/packages"}})
+
+(deftask uranium
+  "Ensure uranium.js and uranium.js.mem are in target directory.  This is not
+  needed to build the package, rather its for any projects using the package."
+  []
+  ;;uranium.js and uranium.js.mem need to both be in the same subdir as b4w's target html.
+  (sift :add-jar {'cljsjs/blend4web #"^*/uranium.js*"}
+        :move {#"^*cljsjs/blend4web/common/uranium.js"
+               "uranium.js"}
+        :to-resource #{#"uranium.js*"}))
+
+(deftask package []
+  (comp
+   (download :url (format "https://raw.githubusercontent.com/TriumphLLC/Blend4Web/%s/deploy/apps/common/b4w.min.js" +lib-version+)
+             :checksum "316e4cb2748cb72bc5a563d1f7fe354c")
+   (download :url (format "https://raw.githubusercontent.com/TriumphLLC/Blend4Web/%s/tools/closure-compiler/extern_fullscreen.js" +lib-version+)
+             :checksum "fe0a688b40a2f0c672e779f0aaced2a8")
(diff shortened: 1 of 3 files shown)


blend4web/build.boot
```diff
@@ -0,0 +1,216 @@
...
+  []
+  ;;uranium.js and uranium.js.mem need to both be in the same subdir as b4w's target html.
+  (sift :add-jar {'cljsjs/blend4web #"^*/uranium.js*"}
+        :move {#"^*cljsjs/blend4web/common/uranium.js"
+               "uranium.js"}
+        :to-resource #{#"uranium.js*"}))
+
+(deftask package []
+  (comp
+   (download :url (format "https://raw.githubusercontent.com/TriumphLLC/Blend4Web/%s/deploy/apps/common/b4w.min.js" +lib-version+)
```
Deraen: Huh, instead of downloading all these invidual files, it is much simpler to just download the archive: `https://github.com/TriumphLLC/Blend4Web/archive/16.11.zip`. Just add `:decompress true` option to download task to also extract the files.

blend4web/build.boot
```diff
@@ -0,0 +1,216 @@
...
+             :checksum "732491e7a8c8db3cbd2fa8ca5f0f0090")
+   (download :url (format "https://raw.githubusercontent.com/TriumphLLC/Blend4Web/%s/src/ext/version.js" +lib-version+)
+             :checksum "ebb10a947f1b609fc168838a73f1357f")
+   (download :url (format "https://raw.githubusercontent.com/TriumphLLC/Blend4Web/%s/deploy/apps/common/uranium.js" +lib-version+)
+             :checksum "9B298C15DAE3D526F9652D896A088CC1")
+   (download :url (format "https://raw.githubusercontent.com/TriumphLLC/Blend4Web/%s/deploy/apps/common/uranium.js.mem" +lib-version+)
+             :checksum "3FF638D920819952240A1FDC28E6FD16")
+   (sift :move {;;These are all the externs
+                #"^b4w.js"
+                "cljsjs/blend4web/common/b4w.ext.js"
```
Deraen: Something like #"^(.*)\.js" "cljsjs/blend4web/common/\1.ext.js" should take care of all these externs. Though it also might match other files.

mikebelanger: Oh sweet - I didn't even know about that `:decompress` flag :P  While I like the idea of a shorter `build.boot`, the zip is 1.4 gigs big.  From a Canadian internet perspective, this means a long download time.  But that's my bias.

Deraen: Woah! Yeah, I can see not wanting to download that file.

mikebelanger: Also, I should mention that `:advanced` does work, contrary to what I posted on slack.  Only drawback is `:advanced` spews a massive dump of warnings (not errors) about useless statements in extern files.  Doesn't appear to affect the output, but looks bad.

Deraen:
I will need to think a bit what is the best solution for this. Downloading 1.4GB file is inconvenient, but so is maintaining packaging script which downloads this many individual files.

mikebelanger:
I'm not a fan of the huge build script either, I just couldn't think of a
better way.

What just occurred to me is trying some kind of iteration for side-effects,
(ie `(doseq)` using the (download... function, maybe thread-first too.  I
mean the root url only changes twice, its just the end filenames that are
different.  A sequence of the filenames would need to be maintained, but
that's it.

Deraen:
Just checked the file contents: files from `src/ext/` are not extern files. Perhaps they are not needed at all?

Deraen:
w3c externs are also not needed, they are already bundled in clousre compiler. Not sure about others. Jquery extern is quite likely unnecessary because if that is needed, this package should just depend on `cljsjs/jquery`.

Deraen:
@mikebelanger I cleaned this as #868, but now I notice one quite large problem: there is no extern file at all for the library itself? The script uses source file as extern, but that is not okay.

mikebelanger:
Thanks again.  The ones in `src/ext` are necessary if you'd like to compile using `:advanced`.  At least, that's been my experience here.  Otherwise a number of things get munged.

Jquery and w3c are likely not important, sure we can probably get rid of them, I'm pretty sure they're only used if you'd like to use make complex games using some kind of persistent state.  Things that are likely better left to more idiomatic clojurescript-libraries.

There is no extern file in the library itself because the externs are already provided by blend4web (it also uses Google Closure).  Most of them appear to be here:

https://github.com/TriumphLLC/Blend4Web/tree/master/tools/closure-compiler

PR #866, opened 2016-12-06
Maptastic

mikebelanger:
New package:

**Extern:** Written by hand


Deraen:
Problems:

- This seems to also include blend4commits, this should be separate to that.
- maptastic js file is included in repo
- unncessary .keep files

mikebelanger:
Oh I'm a fool!  Sorry, would it be better to just do a separate fork of cljsjs altogether?  I don't mind.

Deraen:
Well, it doesn't need to be exactly a separate fork, it it enough to create separate branch in your fork. Then you can create a new PR from that branch.

PR #868, opened 2016-12-06
Mikebelanger blend4web

Deraen:
Replaces #860


Deraen:
@mikebelanger I added a proper generated extern file here. Please test if this works. I have another idea on how to create extern file if the generator doesn't work correctly for this project.

To checkout this branch you can use these commands:

```bash
git fetch origin
git checkout mikebelanger-blend4web
```

mikebelanger:
@Deraen Thanks, but this doesn't work.  I tried installing that on my local maven cache, I then tried compiling [this project](https://github.com/mikebelanger/mic-reactive) and entered `boot deploy`, which compiles using `:advanced`.  Tonejs works (it reports itself booting), but I get a `a.wc is not a function`, likely due to improper munging.  If you'd like to see the output when this works, [check it out here](http://clojurescript-experiments.neocities.org/).  You'll need a microphone.

While I'm all for finding a simpler build script, I don't understand the need to reverse-engineer a set of externs, when there's already externs provided by blend4web.  As far as build script length goes, I might be able to use a `doseq` loop, as the download url doesn't change that much.

Deraen:
> when there's already externs provided by blend4web

Where? Source files are not externs... But as files in `src/ext/` only contain public functions, perhaps it would be okay to use them...

>  I might be able to use a doseq loop, as the download url doesn't change that much

`download` function doesn't do any side-effects itself, it creates the middleware function used by Boot, so `doseq` can't be used here. But you can use `reduce` to build the middleware pipeline.

mikebelanger:
> Where? Source files are not externs... 

My assumption was that the ones in `src/` root are source files, and `src/ext` were (some of) the externs.  When I started this project, I didn't include the ones in `src/ext`, and I couldn't get it to compile in `:advanced`.   Once I included all of those, I was able to compile them in `:advanced`.

> perhaps it would be okay to use them...

It works, but definitely has its drawbacks.  As you have correctly pointed out, the build script is long, and compiling in `:advanced`, while it works, does complain about all the ineffectual local var assignments, which is annoying, and (now that I think about) probably because its not meant as an extern :P

With all that said, I'm not married to any of this, but I can't get it to work any other way.  You mentioned you have another way of generating these externs, and if its simpler/better, I'm all for it.

Deraen:
Updated with the same extern file that blend4web uses during Closure optimization. I modified their build script to output this file, normally the file is removed after build.

I tried this with your project but got an error. But it doesn't seem to be extern related error perhaps?

Deraen:
Hmm, I think I got this working. I changed the uranium task a bit and used serve task to serve the advanced compiled code.

I added uranium task back with rewritten docstring.

PR #869, opened 2016-12-08
Maptastic

mikebelanger:
New package:

**Extern:** Written by hand.


maptastic/build.boot
```diff
@@ -0,0 +1,39 @@
...
+       :description "Javascript/CSS projection mapping utility."
+       :url         "https://github.com/glowbox/maptasticjs"
+       :license     {"Copyright (C) 2015 by Glowbox LLC"
+                     "https://raw.githubusercontent.com/glowbox/maptasticjs/master/license.txt"}
+       :scm         {:url "https://github.com/cljsjs/packages"}})
+
+(deftask package []
+  (comp
+   (download :url
+             "https://raw.githubusercontent.com/glowbox/maptasticjs/master/build/maptastic.min.js"
```
Deraen: Please use specific version or git sha here (if library is not versioned), for reproducible build.

Deraen: And like with versions, please save the sha in a def, like +lib-version+.

Deraen: Like in this package: https://github.com/cljsjs/packages/blob/a486e6af0389a1d2754e3a21e7772f082b208815/emojione-picker/build.boot#L12

mikebelanger: Sure - to be sure you mean the sha of the original repo?

Deraen:
I merged this with some fixes:

- It is important that `+lib-version+` and `+version+` are defined (the names must be these)
- No need to require built-in tasks namespace
- `:version` must use `+version+`, `+version+` must contain the build identifier `-0`
- Fixed indents in package task
- Removed `show` task which did nothing

mikebelanger:
Oops - thanks again!

PR #884, opened 2016-12-15
add rot.js

mohsenil85:
New package: Rot.js, Roguelike toolkit

more info:  https://github.com/ondras/rot.js/

**Extern:** Generated.

### DIFF
--- a/dygraph/README.md
+++ b/dygraph/README.md
@@ -1,7 +1,7 @@
 # cljsjs/dygraph
 [](dependency)
 ```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
Please use versioned url to download the archive or the files. Use the `+lib-version+` var you have defined to build the url:

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

PR #887, opened 2016-12-16
Update README.md

bjconlan:
Update:

Updated namespace in example code.

### DIFF
--- a/inferno/README.md
+++ b/inferno/README.md
@@ -12,7 +12,7 @@ you can require the packaged library like so:
 
 ```clojure
 (ns application.core
-  (:require [cljsjs.inferno]))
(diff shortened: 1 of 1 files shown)


Deraen:
Hmh, in fact `cljsjs.inferno` is the correct practice and the package script is wrong in this case.

Deraen:
@monjohn What do you think? Would it still be possible to rename the namespace provided by Inferno package `cljs.inferno` -> `cljsjs.inferno`.

monjohn:
👍 Yes, thanks for catching my oversight.

Deraen:
https://github.com/cljsjs/packages/commit/a82b9605a843adfd0c9d55f9023a5b9886113603

PR #898, opened 2016-12-29
Package Material Design Components for the web

r0man:

New package:

**Extern:** Written by hand.


Deraen:
I had to revert this as this broke CI builds and I'm not sure of the problem & can't test this locally as I don't have recent enough Node.

r0man:
@Deraen I totally forgot about this problem. I had it on my machine as well. It's a problem with boot and directory symlinks. This should be fixed in boot 2.7.1. Can we run the build with that version?

PR #925, opened 2017-01-10
Added missing ReactDOM dependency

danielglauser:
Update:

ReactDOM has been broken out from the core React library and needs to be loaded as a separate dependency.


Deraen:

PR #926, opened 2017-01-11
Adding react-modal

hlolli:
very small react library by facebook, added externs by hand (based on react-sticky externs), the component works after an advanced compilation, so as far as I see, it works.

New package:
**Extern:** Written by hand.

### DIFF
--- a/react-modal/README.md
+++ b/react-modal/README.md
@@ -1,0 +1,26 @@
+# cljsjs/react-sticky
+
+[](dependency)
+```clojure
+[cljsjs/react-sticky "5.0.4-0"] ;; latest release
+```
+[](/dependency)
+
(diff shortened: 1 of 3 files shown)


ducky427:
I think `react-modal` has a dependency on `react-dom` as well. Also could you squash all the commits into one?

hlolli:
Will do, I did add it initially with [cljsjs/react-dom "15.4.0-0"], had some conflicts but that could have been on my side. In case react-dom bugs someone else, excluding it could potentially solve things since some projects are very selfish on which react-dom package is used (cljs-material-ui is a good example thereof).

hlolli:
pff, don't know much about squashing read this here https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit didn't help much. I can always just refork...

PR #928, opened 2017-01-11
Two new packages: zxcvbn and babel-polyfill

binarykitchen:
Both externs written by hand and tested.

### DIFF
--- a/babel-polyfill/build.boot
+++ b/babel-polyfill/build.boot
@@ -1,0 +1,28 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2" :scope "test"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +lib-version+ "6.20.0")
+(def +version+ (str +lib-version+ "-2"))
+
+(task-options!
+  pom {:project     'cljsjs/babel-polyfill
+       :version     +version+
+       :description "Provides polyfills necessary for a full ES2015+ environment."
+       :url         "https://github.com/babel/babel/tree/master/packages/babel-polyfill"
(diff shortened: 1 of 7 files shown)


babel-polyfill/build.boot
```diff
@@ -0,0 +1,28 @@
...
+  pom {:project     'cljsjs/babel-polyfill
+       :version     +version+
+       :description "Provides polyfills necessary for a full ES2015+ environment."
+       :url         "https://github.com/babel/babel/tree/master/packages/babel-polyfill"
+       :scm         {:url "https://github.com/SMX-LTD/cljsjs-packages"}
+       :license     {"MIT" "https://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+    (download :url "https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/6.20.0/polyfill.js")
```
Deraen: Please use versioned url to download the archive or the files. Use the `+lib-version+` var you have defined to build the url:

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

binarykitchen: @Deraen understand, but i am afraid, not even those files exist on their repo. have a look
https://github.com/babel/babel/tree/master/packages/babel-polyfill

there are no polyfill.js and no polyfill.min.js

Deraen: The CDN has versioned urls: https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/6.20.0/polyfill.js

binarykitchen:
@Deraen oh i get it ... here you go, fixed with versions in URLs for babel-polyfill from CDN

PR #931, opened 2017-01-12
added kjua

MaxSt:
New package:

**Extern:** Generated.

### DIFF
--- a/kjua/build.boot
+++ b/kjua/build.boot
@@ -1,0 +1,30 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2" :scope "test"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +lib-version+ "0.1.1")
+(def +version+ (str +lib-version+ "-0"))
+
+(task-options!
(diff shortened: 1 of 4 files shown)


kjua/build.boot
```diff
@@ -0,0 +1,30 @@
...
+       :scm         {:url "https://github.com/cljsjs/packages"}})
+
+(deftask package []
+  ;;(task-options! push {:ensure-branch nil})
+  (comp
+   (download :url (format "https://github.com/lrsjng/kjua/archive/v%s.zip" +lib-version+)
+             :checksum "79855f89e17867bb7795a61ca9326c46"
+             :unzip true)
+   (sift :move {#"^kjua-[^\/]*/dist/kjua\.min\.js" "cljsjs/development/kjua.inc.js"
+                #"^kjua-[^\/]*/dist/kjua\.min\.js" "cljsjs/production/kjua.inc.js"})
```
Deraen: It is unnecessary to copy the same file to multiple places, and `deps-cljs` won't even pick up the file from production as it doesn't end in `.min.inc.js`.

So, do this:

```
   (sift :move {#"^kjua-[^\/]*/dist/kjua\.min\.js" "cljsjs/common/kjua.inc.js"})
```

PR #933, opened 2017-01-13
js-beautify for html code

hlolli:
This is probably not the ideal solution for a library with multiple js files. The problem is that in current js-beautify there's only beautification for js code while the library offers a beautification for .css and .html in these seperate .js files.

Using :ecmascript5 for minification was important to loose of errors complaining about commas and IE8.

This 1 extern was tested with advanced compilation.

New package:
**Extern:** Written by hand.

### DIFF
--- a/js-beautify/README.md
+++ b/js-beautify/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/js-beautify "1.6.2-0"] ;; latest release
+[cljsjs/js-beautify "1.6.8-0"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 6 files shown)


Deraen:
So does `beautify.js` provide beautify for JS files, `beautify-css.js` for CSS files and `beautify-html.js` for HTML files? Are all these files usable by themselves without others?

Proper way to package this would probably provide three namespaces from this cljsjs package, `deps.cljs` would look something like:

```clj
{:foreign-libs [{:provides ["cljsjs.js-beautify"] :file "cljsjs/js-beautify/beautify.inc.js" :file-min "..."}]
                {:provides ["cljsjs.js-beautify.html"] :file "cljsjs/js-beautify/beautify-html.inc.js" :file-min "..."}
                {:provides ["cljsjs.js-beautify.css"] :file "cljsjs/js-beautify/beautify-css.inc.js" :file-min "..."}]
 :externs ["cljsjs/js-beautify/common/js-beautify.ext.js"]}
```

And copy all the files to proper places.

Deraen:
Oh, and please update the existing package with described deps.cljs.

hlolli:
Sounds good, should even keep backwards compatability as the default would
be js beautifier namespace. I will probably make a new pull request next
monday when im back at my work computer.

hlolli:
Im bit confused by this squashing, I squashed twice since I accedently added another library I want to make pull request for later (but I should have used a branch).

Deraen:
Heh, rebasing can sometimes mess up the author.

I did try to clean this up a bit, but didn't succeed completely myself: https://github.com/cljsjs/packages/commit/97145c92ef5c6dd76bfb7f62cead1bbb5ecaaae9

Anyway, the change is now merged.

PR #940, opened 2017-01-16
[semantic-ui-react] Initial commit v0.64.0

tvanhens:
Only one extern var is set `semanticUIReact` since all react elements can be safely retrieved via `aget` from that class.

New package:

**Extern:** Written by hand.

### DIFF
--- a/semantic-ui-react/README.md
+++ b/semantic-ui-react/README.md
@@ -17,4 +17,30 @@ you can require the packaged library like so:
   (:require cljsjs.semantic-ui-react))
 ```
 
+Currently the only extern declared is `semanticUIReact`. All
+components are available via properties of `js/semanticUIReact` and
+can be accessed via `goog.object/get` and
+`goog.object/getValueByKeys`.
+
+Example:
+
+```clojure
+(ns application.core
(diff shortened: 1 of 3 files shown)


Deraen:
`aget` is only for arrays and must not be used to retrieve properties from objects (https://groups.google.com/forum/#!topic/clojurescript/XJLWNnMR5ik). Instead, one can use `goog.object/get`.

Also, using `goog.object/get` and skipping externs holds true for any library. As the object of Cljsjs is to provide packages with externs, it would be preferable to provide real extern here. Or at the minimum, provide example of accessing properties in the README.

tvanhens:
Thank you for the feedback, wasn't aware of that PSA regarding `aget`. I have elected to update the README.md to show proper usage via `goog.object` access methods.

The reasoning for this is semantic-ui-react is under very active development and new modules are still being added. For the time-being dynamic access via `goog.object` exposes the full functionality of the library without having to update the externs with each new update.

However, I'd be happy to look into the automating the process of generating all the externs but I'd prefer to do than in a future PR/update if possible. I don't think its worth delaying full access to this library via cljsjs.

PR #942, opened 2017-01-16
Add introjs package

lsnape:
New package:

**Extern:** Generated.


introjs/build.boot
```diff
@@ -0,0 +1,29 @@
...
+      :description "Step-by-step guide and feature introduction"
+      :url         "http://introjs.com/"
+      :scm         {:url "https://github.com/usablica/intro.js"}
+      :license     {"GNU AGPLv3" "https://www.gnu.org/licenses/agpl-3.0.en.html"}})
+
+(deftask package []
+  (comp
+   (download :url (str "https://github.com/usablica/intro.js/archive/v" +lib-version+ ".zip")
+             :unzip true)
+   (sift :move {#"intro.js-2.4.0/intro.js"                 "cljsjs/introjs/development/introjs.inc.js"
```
Deraen: Please write these regexes in a way the don't need to be changed when the version is updated,

e.g. #"intro.js-[^/]*/intro.js" should accept anything but forward slashes in place of of the version number.

lsnape: Good spot, thanks. I've relaxed the regex here.

PR #943, opened 2017-01-18
[leaflet] Add Polyline, Polygon, Control, IControl & GeoJson methods

scottlowe:
Leaflet: I updated the extern by hand.

I've installed the generated .jar locally and have run the resulting optimized code in the browser without issue.

### DIFF
--- a/leaflet/README.md
+++ b/leaflet/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/leaflet "0.7.7-6"] ;; latest release
+[cljsjs/leaflet "0.7.7-7"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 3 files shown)


Deraen:
Could you add a comment to start of the extern file to describe what you have added by hand, else there is good chance that the next person updating Leaflet will just generate the file again and remove any customizations.

scottlowe:
@Deraen Looking at the commit history of the Leaflet extern file, it appears to have been added to by hand, over a period of time, so I'm guessing that it was mostly hand-rolled?

I don't mind doing whatever you want to get this fixed, because obviously I want to use this extern; I'm simply wondering if I add comments suggesting that everything else is generated except my stuff, then it could make things worse!

Deraen:
Hmm, you are correct. In this case comment "This extern is written by hand, do not replace with generated file" is enough.

scottlowe:
@Deraen Okay, that's done.

Once these changes are accepted, does a build of downstream packages get triggered, or will I need to submit a version bump for the `react-leaflet` package, also?

Deraen:
Dependents are not automatically updated, if you want to update the, you need to send PRs. Or you can just add the direct dependency to Leaflet to your project to use a specific version.

PR #946, opened 2017-01-20
[web3] Update to version 0.18.2

madvas:
Update:

### DIFF
--- a/web3/build.boot
+++ b/web3/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "0.16.0")
+(def +lib-version+ "0.18.2")
 (def +version+ (str +lib-version+ "-0"))
 
 (task-options!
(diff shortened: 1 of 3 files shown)


web3/build.boot
```diff
@@ -19,9 +19,11 @@
   (comp
     (download :url (str "https://github.com/ethereum/web3.js/archive/" +lib-version+ ".zip")
               :unzip true)    
-    (sift :move {#"web3.js-0.16.0/dist/web3.js" "cljsjs/development/web3.inc.js"
-                 #"web3.js-0.16.0/dist/web3.min.js" "cljsjs/production/web3.min.inc.js"})
-    (sift :include #{#"^cljsjs"})
-    (deps-cljs :name "cljsjs.web3")
+    (sift :move {#"web3.js-0.18.2/dist/web3.js" "cljsjs/development/web3.inc.js"
```
Deraen: Please write these regexes in a way the don't need to be changed when the version is updated,

e.g. #"web3.js-[^/]*/dist/web3.js" should accept anything but forward slashes in place of of the version number.

madvas: yep, okay, fixed, thanks ;)

PR #947, opened 2017-01-20
Inferno

KeeganMyers:
The Inferno library is divided into several packages in the name of modularity. Easy access to several packages in addition to the base Inferno package should simplify development. As such I have added

Inferno-component
  **Extern:** Generated.
Inferno-create-class
    **Extern:** Generated.
Inferno-create-element
    **Extern:** Generated.

I also updated the version number of the base Inferno package in keeping with the latest version.


Deraen:
Okay, so these all are packaged together with `inferno` (based on the CDN url) and all follow the same version number.

I think it would make more sense to package these into the `cljsjs/inferno` package, but as a separate foreign-dependency namespaces. Similar to how js-beautify and other packages provide several namespaces: https://github.com/cljsjs/packages/blob/master/js-beautify/resources/deps.cljs

This way one one package version needs to be updated to update all the files.

KeeganMyers:
That makes sense, I will make that change shortly.

KeeganMyers:
After creating one combined deps.cljs I can't seem to require the namespaces in any other projects. I will re-open this when I resolve the issue.

Deraen:
The `deps.cljs` must be placed on root of classpath, that is, `resources/deps.cljs`.

PR #958, opened 2017-01-26
[react-autosuggest] Update to 8.0.0 version

gigasquid:
Update: react-autosuggest to 8.0.0 version


cldwalker:
Some handy features in this version upgrade. Would be great to get a release on this 😄

PR #961, opened 2017-01-28
[inferno] update to 1.2.2 and add inferno-hyperscript

ul:
Update:

### DIFF
--- a/inferno/README.md
+++ b/inferno/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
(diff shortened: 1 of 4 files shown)


Deraen:
Thanks,

Btw. React package as a example of how to automatically generate md5sums for multiple files: https://github.com/cljsjs/packages/blob/master/react/build.boot#L105-L132

ul:
Oh, interesting. Perhaps we should extract this task into library...

PR #965, opened 2017-01-30
Add flyd

luminusian:
New package:

Extern: Generated.


flyd/build.boot
```diff
@@ -0,0 +1,28 @@
...
+       :url         "https://github.com/paldepind/flyd"
+       :scm         {:url "https://github.com/paldepind/flyd"}
+       :license     {"MIT" "https://github.com/paldepind/flyd/blob/master/LICENSE"}})
+
+(deftask package []
+         (comp
+           (download :url (format "https://github.com/paldepind/flyd/archive/v%s.zip" +lib-version+)
+                     :checksum "eb3e952c894093d0380e629c95cb1e2e"
+                     :unzip true)
+           (sift :move {#"^flyd-0\.2\.3/flyd\.js$" "cljsjs/flyd/development/flyd.inc.js"
```
Deraen: Please use regex that doesn't need to be changed when the package is updated, e.g.

```clj
(sift :move {#"package-[^/]*/lib\.js" "cljsjs/package/development.lib.inc.js"})
```

Where `[^/]*` will allow any number of non slash characters in place of the version number.

PR #974, opened 2017-02-02
Add cljsjs/kld-intersections

arichiardi:
New library
**Extern:** Generated + partly wrtten by hand.

The build fetches all the files it needs to and minifies them in a single `min.inc.js`.


arichiardi:
Hold on on this one, I need to change and link to the license to complete and test it in a project (tomorrow).

intersections/build.boot
```diff
@@ -0,0 +1,113 @@
+(set-env!
+ :repositories #(conj % '["bintray" {:url "http://dl.bintray.com/nitram509/jbrotli"}])
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.6.0" :scope "test"]
+                 [nha/boot-uglify "0.0.7" :scope "test"]])
+
+(def +version+ "1.0.0-0")
```
Deraen: Latest git tag and npm version is 0.1.1

intersections/build.boot
```diff
@@ -0,0 +1,113 @@
...
+
+(task-options!
+ pom  {:project     'cljsjs/intersections
```
Deraen: Package name should be same as in other package manages (npm, bower) or github: "kld-intersections".

intersections/build.boot
```diff
@@ -0,0 +1,113 @@
...
+        (io/make-parents dev-file)
+        (file/copy-atomically prod-file dev-file)
+        (-> fileset (c/add-resource tmp) c/commit!)))))
+
+(deftask package []
+  (let [uglify-dir (normalize-path "uglify/")
+        prod-dir (normalize-path "cljsjs/intersections/production/")
+        dev-dir (normalize-path "cljsjs/intersections/development/")]
+    (comp
+     (download :url "https://raw.githubusercontent.com/thelonious/kld-intersections/development/lib/Intersection.js"
```
Deraen: Please use versioned url to download the archive or the files. Use the `+lib-version+` var you have defined to build the url:

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

intersections/build.boot
```diff
@@ -0,0 +1,113 @@
...
+               :checksum "2FF91BB218A7318DA4F7625A8DCFD27C")
+     (download :url "https://raw.githubusercontent.com/thelonious/kld-intersections/development/lib/IntersectionParams.js"
+               :checksum "6C7979D5AD687F729138DFDE61BD3330")
+     (sift :move {#"^Intersection.js" (str uglify-dir "Intersection.js")})
+     (sift :move {#"^IntersectionParams.js" (str uglify-dir "IntersectionParams.js")})
+
+     ;; dep kld-affine
+     (download :url "https://raw.githubusercontent.com/thelonious/kld-affine/development/lib/Point2D.js"
```
Deraen: It would be preferred to package each dependency as separate package, also, kld-affine has different version than other two packages.

intersections/build.boot
```diff
@@ -0,0 +1,113 @@
...
+               :checksum "BE6181D3332E43A3A74BC7EA53DEFBF6")
+     (download :url "https://raw.githubusercontent.com/thelonious/kld-polynomial/development/lib/PolynomialAdditionalMethods.js"
+               :checksum "A6F11C8AF14580BD88D17214AE2292F9")
+     (download :url "https://raw.githubusercontent.com/thelonious/kld-polynomial/development/lib/SqrtPolynomial.js"
+               :checksum "10CDC2F3F311082711B8A92F6B33300C")
+     (sift :move {#"^Polynomial.js" (str uglify-dir "Polynomial.js")})
+     (sift :move {#"^PolynomialAdditionalMethods.js" (str uglify-dir "PolynomialAdditionalMethods.js")})
+     (sift :move {#"^SqrtPolynomial.js" (str uglify-dir "SqrtPolynomial.js")})
+
+     (uglify :file-name ["Point2D.js"
```
Deraen: I not huge fan of this approach, as the packages also have `index.js` that exposes everything, it should be possible to use browerify/webpack to concat and minify these. But I can also accept this solution.

intersections/build.boot
arichiardi: Ok cool, I will change that

arichiardi:
@Deraen I will answer here on the other requests.
Basically I went to the normal route and was downloading the lastest releases' archives. Unfortunately, they are not in sync with the main repo and nothing was working. I therefore decided to take the files directly from the repo and merge them together as you see above.
This is the only way I could get it working.

intersections/build.boot
arichiardi: This is fair, but I haven't got the time for splitting it and as I say below, it was the only way to make it working...

intersections/build.boot
arichiardi: I can do this, I could use the github SHA if that makes it better.

arichiardi:
If I use the SHA, does it still make sense to use the version number on `npm`? Probably not, shall we name it `0.1.2-SNAPSHOT` or something?

arichiardi:
Thinking a bit more about, in general what a cljsjs package gives me on top of webpack/Bower is that I don't have to use two/three tools and that I have externs handy. Then again, who know what cool things cljs compiler will do for s in the future 😀

Deraen:
There is also this kld-intersections fork with working release with fixes: https://github.com/signavio/svg-intersections. Perhaps that would be possible to package with real version numbers? Dependencies should work with real versions also as that lib is using them from npm.

In my opinion Cljsjs packages should always use real version numbers, that correspond to tags or npm releases, and that the package contents should be original if possible.

> Thinking a bit more about, in general what a cljsjs package gives me on top of webpack/Bower is that I don't have to use two/three tools and that I have externs handy. 

The only thing Cljsjs really provides is automated deployment and the Clojars group id. You can as well use the same tools in your own project and deploy the package with your own group-id.

Deraen:
I tested this library with Common JS module processing and it works: https://clojurescript.org/guides/javascript-modules (at least without advanced optimizations, there is some dynamic code which would require externs in the lib)

arichiardi:
Cool thanks for trying!

Ok so what I can do is to contribute the three modules separately with externs only with the git proper version if you think it is a good option. The unification is due only to the fact that the source has many J's files unless we use browserify/uglify I don't know  of a way to produce a `min.inc.js` for the package.

I was not born as Js developer and I am not super familiar with all the tooling anyways, to any suggestions here is more than welcome 😀

Deraen:
@arichiardi Please try package from https://github.com/cljsjs/packages/pull/979. Browserify seems to work fine for packaging this library for browser use. It uses npm to load the dependencies.

PR #977, opened 2017-02-04
[aws-sdk-js]: fix duplicate keys in aws-sdk-js externs causing `:optimizations "advanced"` to fail on clojurescript 1.9.456

Samstiles:
The current AWS SDK build had 4 duplicate keys in the externs: `createBucket`, `getSignedUrl`, `upload` & `waitFor`. Duplicate keys aren't allowed in ES5 strict mode so as of clojurescript `1.9.456` this cljsjs package does **not** compile with `:optimizations "advanced"`.

See screenshot:

screenshot

This PR removes removes the 4 duplicate keys.


arichiardi:
Yep I had to do the same for `d3`

PR #979, opened 2017-02-05
Arichiardi add intersections

Deraen:
Replaces #974


arichiardi:
Looks good! I will try in my project on Monday. The only thing that I noticed is that externs change a bit and I could not make the [generator](https://github.com/jmmk/javascript-externs-generator) work.

Deraen:
@arichiardi In fact I think most of the externs are unncessary as the package only exposes `intersect` and `shape`: https://github.com/signavio/svg-intersections/blob/master/index.js

arichiardi:
Oh right, the example is a bit different as there is no `Point2D` exposed
anymore...

See on the README: https://github.com/signavio/svg-intersections

Deraen:
Yep, I updated the example to match that.

Deraen:
There is another fork also: https://github.com/Quazistax/kld-intersections but I don't think that has any releases, and it is also using old dependencies. Hopefully fork I packaged works, as I much prefer to package version with proper release.

arichiardi:
Wow quick! :D Thanks a lot for that! The library is very useful and I am
happy we found a solution that is good. So for the feature, should I prefer
shelling out to node for these things?

Deraen:
For creating browser compatible libraries from Common JS modules? Yes, I think browserify is usually good enough solution and doesn't require much configuration. Webpack is good solution when the package provides existing configuration.

arichiardi:
I have tried that, not very up to date, and the maintainer of
`kld-intersections` even sent an email back to me, that is why I picked it
:)

arichiardi:
Yes, cool I am going to close the other one then. Thanks I will follow that
path.

Deraen:
@arichiardi Have you had time to test this?

arichiardi:
Hey Juho, no but soon, I have been stuck on some other task. Sorry about that!
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Deraen:
ping @arichiardi

arichiardi:
Thanks for pinging,. Ihaven't had time yet, sorry, other tasks are taking all my time atm, it is scheduled of course 😀

Deraen:
ping @arichiardi 

I think I tested this myself so I'm thinking about just merging this to get the PR closed :)

arichiardi:
Agree, I think it safely mergeable 😀
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #989, opened 2017-02-10
[chartjs] Bump to 2.5.0 and freshly generated externs

kennethkalmer:
Did it as two separate commits, just in case the whole new externs file isn't appropriate.


Deraen:
Probably 90% of that extern is unnecessary and duplicated code, but if it works there is probably not much reason to worry about it.

kennethkalmer:
I think that might be the case for many generated externs, but I thought rather be safe then sorry.

PR #1004, opened 2017-02-20
[smooth-scroll] Update smooth-scroll version to 10.2.1

macluck:
Updated version to latest


Deraen:
Version 10 removed `escapeChars` from the public api, but doesn't matter that the externs still contain it.

PR #1007, opened 2017-02-21
Added Gun-db

petergarbers:
New package: ( `cljsjs/gun` )
**Extern:** Generated.


petergarbers:
I'm somewhat of a cljs newbie so I apologize if this is done incorrectly

gun/build.boot
```diff
@@ -0,0 +1,28 @@
...
+       :url         "https://github.com/amark/gun"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "https://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+   (download :url (format "https://github.com/amark/gun/archive/v%s.zip" +lib-version+)
+             :checksum "70EE5F8EA4C62B090E4896A96DA40A71"
+             :unzip true)
+   (sift :move {#".*gun\.js$"   "cljsjs/gun/production/gun.inc.js"})
```
petergarbers: This package does not come with a minified version yet. Judging from their github. This should come with the next release

petergarbers:
Gun does not come with a minified file until the next release. As a result I am trying to use the `minify` fn.
It is throwing the following error:

```
Minifying gun.inc.js
Feb 21, 2017 11:02:26 AM com.google.javascript.jscomp.LoggerErrorManager println
SEVERE: /home/peter/.boot/cache/tmp/home/peter/code/cljs-packages/gun/erw/u2qisy/cljsjs/gun/development/gun.inc.js:676: ERROR - Parse error. IE8 (and below) will parse trailing commas in array and object literals incorrectly. If you are targeting newer versions of JS, set the appropriate language_in option.
					lex: (Gun.text.is(lex) || Gun.num.is(lex))? Gun.is.rel.ify(lex) : lex,
					                                                                     ^

Sifting output files...
```
I'm not really sure what to do. Would someone point me in the right direction.
I have looked at specifying the `language_in` option, however literally none of the other cljsjs projects do this. So I'm not sure that it's the correct direction

PR #1014, opened 2017-02-24
Update Tether to 1.3.3

kennyjwilli:
Update:


Deraen:
Please revert the whitespace changes, simpler diff makes it easier for me to check what has changed.

Deraen:
I merged https://github.com/cljsjs/packages/pull/1013 instead, so no need to change this.

PR #1016, opened 2017-02-24
Add reactstrap

kennyjwilli:
New package:

**Extern:** Written by hand.


reactstrap/build.boot
```diff
@@ -0,0 +1,30 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2" :scope "test"]])
```
Deraen: I think this library should depend on `cljsjs/react-with-addons` and `cljsjs/react-dom`.

reactstrap/build.boot
```diff
@@ -0,0 +1,30 @@
...
+      [min?]
+      (format "https://unpkg.com/reactstrap@%s/dist/reactstrap.%sjs" +lib-version+ (if min? "min." "")))
+
+(deftask package []
+         (comp
+           (download :url (download-url false) :name "reactstrap.js")
+           (download :url (download-url true) :name "reactstrap.min.js")
+           (sift :move {#"reactstrap.js"     "cljsjs/reactstrap/development/reactstrap.inc.js"
+                        #"reactstrap.min.js" "cljsjs/reactstrap/production/reactstrap.min.inc.js"})
+           (deps-cljs :name "cljsjs.reactstrap")
```
Deraen: `cljsjs.react` and `cljsjs.react-dom` should be required in `deps.cljs` so ClojureScript can load the files in correct order, so that React is loaded before Reactstrap.

kennyjwilli:
@Deraen I added react-withaddons and react-dom as dependencies and required them in deps.cljs.

kennyjwilli:
Hmm. After adding cljsjs.react and cljsjs.react-dom to `:requires` in the `deps-cljs` task, I now get this error when requiring the lib in my project: 
```
Uncaught Error: Undefined nameToPath for cljsjs.react_dom
```

If I remove `cljsjs.react` and `cljsjs.react-dom` from `:requires` I do not get that error. Am I missing a step here when using `deps-cljs`?

Deraen:
The correct namespace would be `cljsjs.react.dom`: https://github.com/cljsjs/packages/blob/master/react/build.boot#L79

Please test the changes before pushing them here, even if they are suggested by me :)

kennyjwilli:
Yup, my bad. I have pushed three new commits to my branch that should be merged. The first is to exclude `cljsjs/react` from cljsjs/react-dom. If you don't do this you will get React without the addons. Then one fixing the incorrect namespace name. And finally one that removes my note about dependencies in the README that is not needed anymore.

PR #1018, opened 2017-02-26
add iconv-lite

bolasblack:

New package:

**Extern:** Written by hand.

### DIFF
--- a/iconv-lite/.gitignore
+++ b/iconv-lite/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/iconv-lite/README.md
+++ b/iconv-lite/README.md
@@ -1,0 +1,16 @@
+# cljsjs/iconv-lite
+
+```clojure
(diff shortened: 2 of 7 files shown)


Deraen:
What is the yarn.lock file? Does it have to be included in the git?

bolasblack:
@Deraen yarn.lock is generated by [yarn](http://yarnpkg.com/) (a replacement of npm), like Gemfile.lock for bundle. If you npm is enough, we can remove yarn.lock and change the command `yarn` to `npm` in `build.boot`

Deraen:
npm is enough for all other packages so I think it is enough for this. I prefer to keep the packages as similar as possible.

In this case I doubt that npm vs. yarn makes much difference, as iconv-lite itself doesn't have any dependencies, so only thing that might change is transitive dependencies of browserify and uglifyjs.

bolasblack:
@Deraen yarn.lock has deleted now :)

bolasblack:
🍻

PR #1019, opened 2017-02-26
Updated react-onclickoutside to version 5.9.0, 

kovacnica:
Update (react-onclickoutside to version 0.5.9):

Update (react-datepicker to version 0.41.1):

New package (draft-js version 0.10.0):
**Extern:** Generated.


kovacnica:
Hi,

I've built packages that were changed and installed them locally. We are using react-datepicker in production with advanced compilation and for now  everything looks ok with this build. 

Can somebody provide me with a hint why circleCI  is failing? It looks like it is boot/java issue... **User limit of inotify watches reached**

Am I wrong?

Deraen:
Please create separate PRs of each update and new package.

The updates look fine, but draft-js is going to need some changes. React build.boot is not a good example for a new package, it is doing lots of special handling that is unnecessary for all the other packages.

kovacnica:
Thanks... I'll try to simplify draft-js and make separate RPs.

PR #1026, opened 2017-03-01
Add moment timezone package

bleadof:
Add moment timezone

New package:

**Extern:** Written by hand.

### DIFF
--- a/moment-timezone/build.boot
+++ b/moment-timezone/build.boot
@@ -18,11 +18,6 @@
         :license     {"MIT" ""}
         :scm         {:url "https://github.com/cljsjs/packages"}})
 
-(require '[boot.core :as c]
-         '[boot.tmpdir :as tmpd]
-         '[clojure.java.io :as io]
-         '[clojure.string :as string])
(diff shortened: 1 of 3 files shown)


bleadof:
Ugh, actually did this a while back, so it has old versions. I'll see if they upgrade nicely...

moment-timezone/build.boot
```diff
@@ -0,0 +1,40 @@
...
+        :version     +version+
+        :description "A javascript date library for parsing, validating, manipulating, and formatting dates."
+        :url         "http://momentjs.com/timezone"
+        :license     {"MIT" ""}
+        :scm         {:url "https://github.com/cljsjs/packages"}})
+
+(require '[boot.core :as c]
+         '[boot.tmpdir :as tmpd]
+         '[clojure.java.io :as io]
+         '[clojure.string :as string])
```
Deraen: Please remove these unncessary requires.

bleadof: Done

PR #1033, opened 2017-03-10
react-vis-force 0.1.4

JoelSanchez:
New package:
**Extern:** Generated.

The library works but I don't know how to handle the JS map and the .css and .css.map files.


Deraen:
You can copy the CSS file to `cljsjs/react-vis-force/common` folder (preferable all files should be under `cljsjs/<package-name>/...`).

Cljsjs doesn't support source-maps for foreign-libs and neither do any CSS preprocessors I know about so it is not necessary to copy `.js.map` or `.css.map` files.

react-vis-force/build.boot
```diff
@@ -0,0 +1,32 @@
...
+
+(require '[clojure.java.io :as io])
+
+(deftask package []
+  (comp
+    (download :url (str "https://unpkg.com/react-vis-force@" +lib-version+ "/dist/react-vis-force.js")
+              :checksum "5860feef46d67e9b73a3171c0ea51294")
+    (download :url (str "https://unpkg.com/react-vis-force@" +lib-version+ "/dist/react-vis-force.min.js")
+              :checksum "b7fbfc6ecbabef2dc68c995c040f1c80")
+    (sift :move {#"^react-vis-force.js"      "cljsjs/development/react-vis-force.inc.js"
```
Deraen: Add package name to the path: `cljsjs/react-vis-force/development/react-vis-force.inc.js`.

Deraen: Download the CSS file similar to JS files and drop `.js.map` and `.css.map` files from repo.

react-vis-force/build.boot
```diff
@@ -0,0 +1,32 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2" :scope "test"]])
```
Deraen: In addition to `deps-cljs :require` package must also depend on `cljsjs/react` if React is needed, so add a dependency here.

JoelSanchez:
Applied the requested changes.

PR #1038, opened 2017-03-13
React onclickoutside

kovacnica:
Update:


Deraen:
How did you check that the API didn't change? Looking at the repository, v5 seems to have been quite large change.

New tests for example mention method `handleClickOutside` which is not included in the externs.

Deraen:
Also, this branch seems to contain some conflicts in openlayers and material-components files? Perhaps the branch is based on old commit?

PR #1062, opened 2017-03-28
Bump react-dates to 8.2.1; add a few bits of documentation.

ddellacosta:
~~Soooo I tried generating externs automatically via https://github.com/jmmk/javascript-externs-generator, but it gave me some nasty output:~~
```<snip>```
~~I'm going to keep trying to figure it out but if you have suggestions let me know. Ideally, I'd like to avoid writing all these externs by hand.~~

I've updated the externs file, let me know if there's anything else you want here.

### DIFF
--- a/react-dates/README.md
+++ b/react-dates/README.md
@@ -1,8 +1,12 @@
 # cljsjs/react-dates
 
+This wraps up [Airbnb's react-dates widget](https://github.com/airbnb/react-dates) in a tasty CLJSJS wrapper.
+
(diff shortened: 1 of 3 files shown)


ddellacosta:
Many thanks to @jmmk for [showing me how to generate the extern file](https://github.com/jmmk/javascript-externs-generator/issues/32).

ddellacosta:
Looks like the build simply timed out--I'd restart it but I don't have permissions.

PR #1068, opened 2017-03-29
Create package for fela-plugin-prefixer.js

derHowie:
New package: fela-plugin-prefixer.js

**Extern:** Written by hand.

### DIFF
--- a/fela-plugin-prefixer/README.md
+++ b/fela-plugin-prefixer/README.md
@@ -1,0 +1,22 @@
+# cljsjs/fela-plugin-prefixer
+
+[](dependency)
+```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
In future, it might make sense to package all the fela packages in single cljsjs package, as they all seem to follow the same versioning, similar to inferno: https://github.com/cljsjs/packages/blob/master/inferno/resources/deps.cljs

derHowie:
I intend to add some more and would not mind making that change. Does JS packaged in this manner benefit from dead code elimination when compiled?

Deraen:
Yes. (Though not the Dead code elimination specifically, that is done by Closure.) ClojureScript only included those foreign-lib JS files in output that are required by some namespaces.

PR #1071, opened 2017-03-31
add react-popper to cljsjs

hlolli:
New package:
**Extern:** Generated.


hlolli:
ah f***, I didn't branch these two, that were supposed to be two seperate pull requests for two different libraries. If this is a problem I can correct this on monday.

Deraen:
It is fine, they don't depend on each other and neither needs fixes.

PR #1072, opened 2017-03-31
Add XRegExp 3.1.1.

daemianmack:
New package:

**Extern:** Generated.


xregexp/build.boot
```diff
@@ -0,0 +1,36 @@
...
+       :version     +version+
+       :description "Extended JavaScript regular expressions"
+       :url         "https://github.com/slevithan/xregexp"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+
+(deftask package []
+  (comp
+   (download :url "https://raw.githubusercontent.com/slevithan/xregexp/37413619ea9f03638bfa92e038e4b4c15645d446/src/xregexp.js"
```
Deraen: Please use `(format "https://raw.githubusercontent.com/slevithan/xregexp/v%s/src/xregexp.js" +lib-version+)` to build the url here. This way url doesn't need to be changed when version is updated.

daemianmack:
Ah, missed that -- thanks!

xregexp/xregexp.js
```diff
@@ -0,0 +1,1834 @@
+/*!
```
Deraen: This file should not be included in the repo?

daemianmack: True -- fixed.

PR #1083, opened 2017-04-06
[GraphiQL] Add GraphiQL 0.9.3

noonian:
New package:

**Extern:** Written by hand.


graphiql/build.boot
```diff
@@ -0,0 +1,45 @@
+(set-env!
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.5.2" :scope "test"]])
```
Deraen: Package should also depend on `cljsjs/react` package.

noonian:
Thanks. I've added the dependency, squashed the extra commit, and rebased against master.

PR #1086, opened 2017-04-06
Add react-xmasonry

yeehaa123:

New package:

**Extern:** Written by hand.

### DIFF
--- a/react-xmasonry/build.boot
+++ b/react-xmasonry/build.boot
@@ -6,7 +6,7 @@ l  :resource-paths #{"resources"}
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
 (def +lib-version+ "1.0.2")
-(def +version+ (str +lib-version+ "-0"))
+(def +version+ (str +lib-version+ "-1"))
 
 (task-options!
   pom  {:project     'cljsjs/react-xmasonry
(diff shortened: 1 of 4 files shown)


yeehaa123:
@Deraen Okay, I'll add the react package as a dependency. Does not that package already provide the foreign lib require though?

react-xmasonry/build.boot
```diff
@@ -0,0 +1,31 @@
...
+  (comp
+   (download :url (format "https://github.com/ZitRos/react-xmasonry/archive/master.zip")
+             :checksum "E6B128183607A37A561E58E5D913CF4C"
+             :unzip true)
+
+   (sift :move {#"^react-xmasonry-.*/dist/index.js"  "cljsjs/react-xmasonry/common/react-xmasonry.inc.js"})
+
+   (sift :include #{#"^cljsjs"})
+
+   (deps-cljs :name "cljsjs.react-xmasonry")
```
Deraen: The foreign-lib declaration in deps.cljs should require `cljsjs.react` so that React is loaded correctly before this lib: like here https://github.com/cljsjs/packages/blob/master/react-flexbox-grid/build.boot#L71

Cljs docs further document foreign lib requires: https://clojurescript.org/reference/packaging-foreign-deps

yeehaa123:
Okay, Thanks! Will do!

PR #1089, opened 2017-04-08
[react-flexbox-grid] Update to version 1.0.0

madvas:


react-flexbox-grid/build.boot
```diff
@@ -2,8 +2,8 @@
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]])
```
Deraen: This package seems to be missing dependency on `cljsjs/react`.

madvas:
okay fixed

PR #1091, opened 2017-04-09
Updated Pixi to 4.4.3

saftacatalinmihai:
Update:


pixi/build.boot
```diff
@@ -17,11 +17,9 @@
 
 (deftask package []
   (comp
-    (download :url (format "https://github.com/GoodBoyDigital/pixi.js/archive/v%s.zip" +lib-version+)
-              :checksum "24e7926495a8968098b23ebad06f4534"
-              :unzip true)
-    (sift :move {#"^pixi\.js-([\d\.]*)/bin/pixi\.js$" "cljsjs/pixi/development/pixi.inc.js"
-                 #"^pixi\.js-([\d\.]*)/bin/pixi\.min\.js$" "cljsjs/pixi/production/pixi.min.inc.js"})
+    (download :url (str "https://cdnjs.cloudflare.com/ajax/libs/pixi.js/" +lib-version+ "/pixi.min.js")
```
Deraen: As the previous version included both the minified and non-minified file, I'd prefer if the new version also provided both. Non minified file is available at cdnjs and you can just add second `download` file to download that.

PR #1096, opened 2017-04-13
[pikaday] Update to version 1.5.1

laurio:

### DIFF
--- a/pikaday/build.boot
+++ b/pikaday/build.boot
@@ -5,7 +5,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "1.4.0")
+(def +lib-version+ "1.5.1")
 (def +version+ (str +lib-version+ "-1"))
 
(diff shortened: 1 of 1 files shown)


laurio:
Resolves #1097

Deraen:
This version forces to use Moment.

There was two foreign-libs previously so user could select either `cljsjs.pikaday` and `cljsjs.pikaday.with-moment`, where the only difference was that the latter had dependency on moment so that moment js got evaluated before Pikaday. I'm not sure if this is necessary. Perhaps we could just have single foreign lib without moment require, and user can add that to their own project if they want to use Pikaday with Moment?

laurio:
Yes, providing a single dependency without requiring Moment would be better imo. I updated the pull request.

laurio:
Actually the original issue was caused by a regression in Clojurescript v1.9.518 - i reverted changes in `pikaday/resources/deps.cljs` file now that the issue was fixed in latest Clojurescript release.

PR #1100, opened 2017-04-17
remove duplicate key in externs

yeehaa123:

### DIFF
--- a/react-xmasonry/build.boot
+++ b/react-xmasonry/build.boot
@@ -6,7 +6,7 @@
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
 (def +lib-version+ "1.0.2")
-(def +version+ (str +lib-version+ "-1"))
+(def +version+ (str +lib-version+ "-2"))
 
 (task-options!
   pom  {:project     'cljsjs/react-xmasonry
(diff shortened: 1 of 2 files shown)


yeehaa123:
Forgot to bump. Should be okay now

react-xmasonry/build.boot
```diff
@@ -19,7 +19,7 @@
 (deftask package []
   (comp
    (download :url (format "https://github.com/ZitRos/react-xmasonry/archive/master.zip")
-             :checksum "E6B128183607A37A561E58E5D913CF4C"
+             :checksum "D9F26537EB811BC06717D08A93E58235"
```
Deraen: The url should use `+lib-version+` and the checksum should stay the same if version is not updated. Now the master branch points to version 2.0.1 instead of 1!

```
(format "https://unpkg.com/react-xmasonry@%s/dist/index.js" +lib-version)
```

Or something, as the package doesn't use git tags.

yeehaa123: You're right. Forgot to update the lib-version. Should be okay now. Sorry.

Deraen: Even still, you should also fix the download url to use `+lib-version+` so that the build is reproducible even if library is updated.

yeehaa123:
Thanks! The library does not use version tags, so I made the download depend on the commit-sha instead.

Deraen:
Like I said in the first commit, if the lib doesn't use git tags, just download the file from npm.

yeehaa123:
I am sorry, but I don't know how to do that. Can you point me to another build.boot file that does that?

Deraen:
https://github.com/cljsjs/packages/blob/8c0413ca69d8d15203e75bc5bff054bf70edf06a/slate/build.boot#L21

for example https://unpkg.com/react-xmasonry@2.0.1/dist/index.js

unpkg is the npm cdn.

yeehaa123:
Awesome! Thanks! Hope this does it....

PR #1101, opened 2017-04-17
Added THREE.DecalGeometry

bmatcuk:
New package:

**Extern:** Written by hand.

### DIFF
--- a/three-decalgeometry/build.boot
+++ b/three-decalgeometry/build.boot
@@ -4,8 +4,11 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
+;; The library doesn't use any kind of versioning scheme, so a sha is used to
+;; guarantee a specific version.
 (def +lib-version+ "0.1.0")
 (def +version+ (str +lib-version+ "-0"))
+(def sha "546cebafebd34b4b10cc47c2d6412310e6372159")
 
 (task-options!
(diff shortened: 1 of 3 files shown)


three-decalgeometry/build.boot
```diff
@@ -0,0 +1,30 @@
...
+ pom {:project     'cljsjs/three-decalgeometry
+      :version     +version+
+      :description "Decal Geometry for three.js"
+      :url         "https://github.com/spite/THREE.DecalGeometry"
+      :scm         {:url "https://github.com/spite/THREE.DecalGeometry"}
+      :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+   (download  :url      "https://raw.githubusercontent.com/spite/THREE.DecalGeometry/546cebafebd34b4b10cc47c2d6412310e6372159/src/THREE.DecalGeometry.js"
```
Deraen: If you absolutely have to use sha1 to download lib, at least define the sha in a var next to lib version and add a note about having to update both when updating package... Though as this lib doesn't use tags and doesn't have a version number, it looks like this is not meant for global use.

bmatcuk:
Ya, the library doesn't have any kind of versioning scheme, so I wanted to use a sha to point to a specific commit to ensure consistency every time someone installs this cljsjs package.

PR #1105, opened 2017-04-20
upgrade to leaflet 1.0.3

sonwh98:
upgrade to leaflet 1.0.3

### DIFF
--- a/leaflet/build.boot
+++ b/leaflet/build.boot
@@ -4,8 +4,8 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "0.7.7")
-(def +version+ (str +lib-version+ "-8"))
+(def +lib-version+ "1.0.3")
+(def +version+ (str +lib-version+ "-1"))
 
 (task-options!
  pom  {:project     'cljsjs/leaflet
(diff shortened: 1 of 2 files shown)


sonwh98:
When I used the current api, it worked for my use cases: add map, add markers, add popups. So it seems the new API only adds and not deletes in which case just upgrading to uses the latest library should do no harm. This is better than using a pre 1.0 library thats a few years old

Deraen:
I'd rather not update the library if I know that the externs are missing stuff. While (old) API works with the existing externs, using any of new methods would break when using advanced compilation.

I'm using Leaflet myself and I have a plan on how to get proper externs for Leaflet and make the updates easier in future. I'll try to implement this soon.

sonwh98:
cool.  how will you make it easier? What I usually do is use API and if something break, create externs for it

Deraen:
Haven't had time to work on this so I'll just merge this, and try to improve this later.

sonwh98:
I've been using deploying to my own private maven repo.  I've made some additions because there were some functions that failed in advance compilation. should I make pull request for my changes or continue to deploy to my private repo and wait for u to make the big changes?  

The way I work is use the api and if it breaks in advance compilation, add externs. I don't create externs for all function in one go because there are too many and error prone. I just create exeterns for the ones I need. that way I can test it too.

Deraen:
Just make PR.

PR #1108, opened 2017-04-23
P5

pablinos:
The externs for this were generated and then hand edited. I've included a file the util wrapper functions which I wasn't sure if it was appropriate. They're not complete for a start!

I've tested that it works with advanced compilation.

New package:

**Extern:** Generated.
**Extern:** Written by hand.


Deraen:
While the util namespaces can be very useful useful, maintaining them here would add to the noise in this repo, which I'd like to keep to minimum. If you want to maintain cljs wrapper for the library, I'd suggest creating own project for it, and linking to it from the readme here.

pablinos:
Ah ok, I did wonder about that! Shall I remove it and create another PR?

Deraen:
Yes, please. (No need for new PR, just push changes to the same branch and PR is updated).

pablinos:
Of course! Ok, I'll get on that shortly.

PR #1111, opened 2017-04-24
Add epubjs v0.2.15

zarkone:
New package:

**Extern:** Generated.


zarkone:
Found typo in pdfjs package and decided to ship it with this PR too, but if it's not ok, I could separate it out to it's own PR.

PR #1116, opened 2017-04-25
New vega

sorenmacbeth:
Updates to vega 3.0 beta, vega-lite 2.0 beta

New package vega-embed

Update:


sorenmacbeth:
The error looks like a sequencing problem. Should I perhaps make a separate PR for each package? I'm guessing yes.

Deraen:
Yes, it would be preferable to separate vega-embed which depends on the updated vega packages to it's own PR. Though it will continue to fail until the updates are merged.

@miikka What do you think about updating vega & vega-lite to beta versions?

What about replacing the hard written extern for vega with generated one, I'm a bit concerned about that. Vega-lite already uses generated extern so that is probably fine.

sorenmacbeth:
@Deraen the generated extern is more complete and there are a lot of new externs for vega 3.0 so I'd much prefer we use there generated one. I will close this PR and submit individual ones. I will do vega first. If we can get that one published, then I'll do PRs for the others.

PR #1117, opened 2017-04-25
vega 3.0 beta

sorenmacbeth:
Bump to Vega 3.0 beta

Update:

This auto-generated extern is more complete and I'd prefer we use it for vega 3.0


miikka:
I was asked in #1116 what I think about upgrading Vega to a beta version. I think it's a good idea. My understanding is that the development on 2.6 branch has ended and the final version of 3.0.0 is going to be ready soon soon. It's a major breaking change and thus I'd recommend using the 3.0 betas instead of 2.6 for any new development.

PR #1128, opened 2017-04-30
Add the Trello Client library

rorygibson:
New package:

**Extern:** Written by hand.

### DIFF
--- a/trello-client/README.md
+++ b/trello-client/README.md
@@ -7,12 +7,10 @@
 [](/dependency)
 
 This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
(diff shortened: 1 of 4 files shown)


Deraen:
Packaging files which are supposed to be loaded directly from the service provider is problematic. If Trello updates the file the file in the package would be out-dated and there is no proper versioning scheme.

I think Trello lib would be best packaged as extern only, similar to Google Maps and a few other libs: https://github.com/cljsjs/packages/wiki/Extern-only-packages

rorygibson:
Yeah, I didn't consider that @Deraen , sorry.

I've updated the PR, could you take another look please? 🙏

PR #1130, opened 2017-05-01
Update build.boot

ghostandthemachine:
Bump version and checksum for new version

Update:


Deraen:
This package should have dependency and foreign library dependency on React and React-DOM: https://github.com/cljsjs/packages/wiki/Foreign-library-dependencies

ghostandthemachine:
Thanks for the feedback. Should be good to go now.

Deraen:
Looks like the latest commit accidentally reverted the version change.

Deraen:
^ ping @ghostandthemachine

PR #1131, opened 2017-05-01
Google diff match patch/20121119 2

fredZen:
Update:

My main goal here was to add the minified JavaScript, but I also hand-extracted the ext file from the library because the existing one was missing a few symbols


fredZen:
Thanks!

PR #1134, opened 2017-05-02
Highcharts (style by css)

danielstockton:
This is the _styled mode_ version of Highcharts:

https://www.highcharts.com/docs/chart-design-and-style/style-by-css

New package:

**Extern:** Written by hand.


Deraen:
I guess the css files should also be included in this version: https://github.com/cljsjs/packages/wiki/Non-JS-Assets

danielstockton:
Are there any existing packages that include non-js assets I can look at?
The css just provides some default styles, it is not a hard dependency.

Deraen:
https://github.com/cljsjs/packages/blob/master/pikaday/build.boot#L26-L27

PR #1136, opened 2017-05-02
[material-ui-chip-input] Update to version 0.13.5

madvas:

### DIFF
--- a/material-ui-chip-input/build.boot
+++ b/material-ui-chip-input/build.boot
@@ -1,8 +1,9 @@
 (set-env!
   :resource-paths #{"resources"}
-  :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]])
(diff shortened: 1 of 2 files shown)


material-ui-chip-input/build.boot
```diff
@@ -2,7 +2,7 @@
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]])
```
Deraen: This package should depend on`cljsjs/react` and `cljsjs/material-ui`.

madvas: okay, fixed ;)

Deraen: Hmm, I forgot that material-ui comes with its own React? If that is still the case, `cljsjs/react` should be dependency here.

madvas: yep, fixed ;)

PR #1141, opened 2017-05-04
Update material components

flyboarder:
Update to material components.

No changes to the API.

Depends on https://github.com/boot-clj/boot/pull/610


Deraen:
I already merged https://github.com/cljsjs/packages/commit/7bf9e71acf2ba3ac973c183d87326f80126071e9 which also had updated externs.

Will something in the updated package cause problems with Boot?

flyboarder:
Yes! This package uses lernt and symlinks which are broken when using add-resource.

Unless you are able to build this package yourself? It fails locally for me.

Deraen:
Seems to have been successful: https://clojars.org/cljsjs/material-components

PR #1144, opened 2017-05-04
init cljsjs.pixi-sound

sniperliu:
cljsjs.pixi-sound with minimum externs used in our project like play, add, find etc.

New package:

**Extern:** Written by hand.

btw: tried automatic generation but failed.

### DIFF
--- a/pixi-sound/build.boot
+++ b/pixi-sound/build.boot
@@ -1,0 +1,29 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]
+                  [cljsjs/pixi "4.4.3-0"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +lib-version+ "1.4.1")
+(def +version+ (str +lib-version+ "-0"))
+
+(task-options!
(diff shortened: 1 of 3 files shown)


pixi-sound/build.boot
```diff
@@ -0,0 +1,29 @@
...
+ pom  {:project     'cljsjs/pixi-sound
+       :version     +version+
+       :description "WebAudio API playback without any Flash shims or HTML Audio fallback."
+       :url         "https://github.com/pixijs/pixi-sound"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+   (download :url  "https://pixijs.github.io/pixi-sound/dist/pixi-sound.js")
```
Deraen: Please use versioned url to download the archive or the files. Use the `+lib-version+` var you have defined to build the url:

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

sniperliu:
1. Made a change to use npm install, instead of downloading from url (could not find pixi-sound versioned url).
1. Add one line in README about limited exposure of api.

Deraen:
Many packages use unpkg to download files from npm. But this approach is fine also.

PR #1145, opened 2017-05-05
Version bump firebase to 3.7.3

lsenta:
Use a variable for the checksum since it changes as often as the
version.

Update:


lsenta:
I've been using it for the last few months, tested with the database only but everything seem's to be working. Please have another user check it out before merging.

PR #1167, opened 2017-05-19
[quill] update to 1.2.4 version

szobov:
Update:

### DIFF
--- a/quill/build.boot
+++ b/quill/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "1.1.0")
+(def +lib-version+ "1.2.4") ;; released at Apr 18, 2017
 (def +version+ (str +lib-version+ "-3"))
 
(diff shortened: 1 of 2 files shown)


Deraen:
Based on the changelog, the API has changed: https://github.com/quilljs/quill/blob/develop/CHANGELOG.md#120

szobov:
But it's only experimental APIs changes and it doesn't break normal API. I should add it to extern?

Deraen:
All the new calls should be added to extern, no matter if experimental or not, else someone will try to use them and they suddenly break in production build.

Deraen:
@Szczyp Add `find`, `getIndex`, `getLeaf`, `getLine`, `getLines` to `externs` and I can merge this.

Szczyp:
@Deraen I feel like I shouldn't be here.

Deraen:
Oops, sorry, meant @szobov. Just typed sz and used autocomplete.

szobov:
@Deraen done!

PR #1171, opened 2017-05-20
[react-transition-group] Add react-transition-group

anmonteiro:
**Extern:** Generated.


Deraen:
Is there a benefit in building the file here instead of using the provided one: https://unpkg.com/react-transition-group@1.1.3/dist/react-transition-group.js ?

anmonteiro:
We can use that one for minified JS, but what about the development version?

Deraen:
Unminified version is not necessary. I think I'd rather just use the official file for both dev and prod. Or do you think there are benefits in having dev version? Proper line numbers for exceptions?

This file looks like unminified UMD build, but I guess it is not being updated: https://github.com/reactjs/react-transition-group/blob/master/index.js

PR #1175, opened 2017-05-26
[react-sticky] Bump to 5.0.8

kennethkalmer:
Also updated the externs and added a comment on how it was generated


kennethkalmer:
I would love to bump it to 6.0.1, but the author removed the amd builds from his webpack config and I don't have the will-power or the time right now to get it done. Will take another stab over the weekend.

PR #1183, opened 2017-06-02
Adding a currency formatting component

sri-ds:

New package: react-number-format

**Extern:** Written by hand.

Update:

### DIFF
--- a/react-number-format/build.boot
+++ b/react-number-format/build.boot
@@ -1,0 +1,38 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]
+                  [cljsjs/react "15.3.1-0"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +lib-version+ "2.0.0")
+(def +version+ (str +lib-version+ "-alpha2"))
+
+(task-options!
+ pom  {:project     'cljsjs/react-number-format
+       :version     +version+
+       :description "Yet another react component for input masking"
+       :url         "https://github.com/s-yadav/react-number-format"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(require '[boot.core :as c]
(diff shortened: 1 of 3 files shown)


martinklepsch:
Closure library also provides `goog.i18n.NumberFormat` which probably does similar things like this: https://google.github.io/closure-library/api/goog.i18n.NumberFormat.html

Doesn't hurt to merge new packages like this but generally using Closure library has some nice benefits (dead code elimination, no dependency).

react-number-format/build.boot
```diff
@@ -0,0 +1,38 @@
...
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(require '[boot.core :as c]
+         '[boot.tmpdir :as tmpd]
+         '[clojure.java.io :as io]
+         '[clojure.string :as string])
+
+(deftask package []
+  (comp
+   (download :url (str "https://github.com/s-yadav/react-number-format/archive/v2.0.0-alpha2.zip")
```
Deraen: Use the `+lib-version+` var you have defined to build the url:

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

react-number-format/build.boot
```diff
@@ -0,0 +1,38 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]
+                  [cljsjs/react "15.3.1-0"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +lib-version+ "2.0.0")
```
Deraen: Use "2.0.0-alpha2" as the lib version and append just -0 as the build identifier to version.

react-number-format/build.boot
sri-ds: Just now made the changes you requested.

sri-ds: @martinklepsch This is a text input component that allows for formatting currency as the user  enters in the text field.

sri-ds: @Deraen Have made the changes you requested. Can you look at this pr again to see if this can be merged? Thanks.

PR #1187, opened 2017-06-03
[fela-preset-dev] Add 4.3.5

voxlet:
New package:

**Extern:** Written by hand.

### DIFF
--- a/fela-preset-dev/README.md
+++ b/fela-preset-dev/README.md
@@ -1,0 +1,22 @@
+# cljsjs/fela-preset-dev
+
+[](dependency)
+```clojure
+[cljsjs/fela-preset-dev "4.3.5-0"] ;; latest release
+```
+[](/dependency)
+
+After adding the above dependency to your project you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require
+    cljsjs.fela
+    cljsjs.fela-preset-dev))
+```
+
+# Example
+
+```clojure
+(.createRenderer js/Fela #js {:plugins js/FelaPresetDev})
+```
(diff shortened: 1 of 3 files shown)


masarito-gy:
This actually doesn't have a hard dependency on `fela` unlike `react-fela`, same as the [fela-plugin-prefixer](https://github.com/cljsjs/packages/blob/master/fela-plugin-prefixer/build.boot) package this one is based off. 

It's an aggregation of these plugins:
- [fela-plugin-extend](https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-extend/package.json)
- [fela-plugin-isolation](https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-isolation/package.json)
- [fela-plugin-lvha](https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-lvha/package.json)
- [fela-plugin-prefixer](https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-prefixer/package.json)
- [fela-plugin-fallback-value](https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-fallback-value/package.json)
- [fela-plugin-unit](https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-unit/package.json)

I double-checked their `package.json` files for dependencies on `fela` and were clean too.

I'm not familiar with the practice in CLJSJS though, if you think it's better to include the dependencies, ping me again and I will add them!

Deraen:
Okay, I think I only checked the react package for dependencies and presumed the others would also depend on fela.

voxlet:
Good to know! (And excuse me for the account mix-up, it's me... :)

PR #1189, opened 2017-06-04
[react] Add extern for createElement

voxlet:
Update:

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -1,7 +1,7 @@
 # cljsjs/react
 
 ```clojure
-[cljsjs/react "15.4.2-2"] ;; latest release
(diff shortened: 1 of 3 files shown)


Deraen:
Did you hit problems without this?

I started how could we have missed this and I think it is because there is extern for DOM `Document.createElement` already so the calls don't get mangled. Anyway, no harm adding this here also.

--- 

masarito-gy:
@Deraen Sorry I didn't notice the build number, also bumped the version in README while I was at it.

No actual problems without the extern, for me this was mostly for the nicety of having Cursive recognize `js/React.createElement`. Interesting point on being lucky with `Document.createElement` though! It does feel somewhat safer with this in place.

PR #1190, opened 2017-06-05
[quill] bump version, remove quill.core.js

minikomi:
* Updated the quill version to 1.2.5
* `quill.core.js` is a minimal version intended for use [with webpack](https://quilljs.com/guides/adding-quill-to-your-build-pipeline/) for a customized build. The target filename `quill.core.inc.js` was coming before `quill.inc.js` when sorted, and during dev only the minimal version was being loaded without themes / modules.

Update:

### DIFF
--- a/quill/build.boot
+++ b/quill/build.boot
@@ -4,8 +4,8 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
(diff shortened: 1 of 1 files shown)


Deraen:
Quill.core.js is UMD module so it can be loaded in a browser.

But having multiple `.inc.js` files with `deps-cljs` task doesn't make much sense.

PR #1191, opened 2017-06-05
[react-bootstrap] Update to 0.31.0

gws:

I also detailed (and updated) the extern process I used in the extern file comments.


react-bootstrap/README.md
```diff
@@ -15,10 +15,10 @@ you can require the packaged library like so:
   (:require cljsjs.react-bootstrap))
 ```
 
-To use this with Boot require "boot-less" like so:
+To use this with Boot, require [less4clj][less4clj] like so:
 ```clojure
 (set-env!
-  :dependencies '[[deraen/boot-less "0.3.0" :scope "test"]
+  :dependencies '[[deraen/less4clj "0.6.2" :scope "test"]
```
Deraen: boot-less is correct in fact, less4clj is just the project and library package, boot-less provides the boot task.

gws: Hey, sorry about that! I think I was thrown by the deprecation warning on the older repo, but I should have been more careful here.

I've changed the link text and dependency pointer back to `boot-less`.

PR #1193, opened 2017-06-06
[react-day-picker] Update to 5.5.3

deepxg:


react-day-picker/build.boot
```diff
@@ -20,16 +20,16 @@
...
   (task-options! push {:ensure-branch nil :tag false})
   (comp
-    (download :url (str "https://raw.githubusercontent.com/gpbl/react-day-picker/v" +lib-version+ "/dist/DayPicker.js")
-              :checksum "0F90D0E05603015BE9C41153A986F38D")
-    (download :url (str "https://raw.githubusercontent.com/gpbl/react-day-picker/v" +lib-version+ "/src/style.css")
-              :checksum "4CEDAFC0F76D42F8574B6240374CE7DA")
-    (sift :move {#"^DayPicker.js$"
-                 "cljsjs/react-day-picker/development/react-day-picker.inc.js"
+   (download :url (str "https://raw.githubusercontent.com/gpbl/react-day-picker/v" +lib-version+ "/lib/daypicker.js")
+             :checksum "08293a1d77ef3b201f6b6e8ea994d465")
```
Deraen: There is minified file also available so load that also: https://github.com/gpbl/react-day-picker/blob/master/lib/daypicker.min.js

And sift that file to path ` "cljsjs/react-day-picker/production/react-day-picker.min.inc.js"`

deepxg: Okay, I have added the minified JS, take a look.

PR #1198, opened 2017-06-07
[react-day-picker] Add additional API externs.

deepxg:


react-day-picker/resources/cljsjs/react-day-picker/react-day-picker.ext.js
```diff
@@ -543,7 +543,30 @@ var DayPicker = {
 DayPicker.prototype = {
   "forceUpdate": function () {},
   "isReactComponent": function () {},
-  "setState": function () {}
+  "setState": function () {},
```
Deraen: So these are added by hand? Please add a comment at the top of the file so one who updates this the next time knows the generated externs need to be modified, and add a comment here, something like `// methods below are added by hand`.

PR #1200, opened 2017-06-09
[pixi] Upgrade Pixi to 4.5.2

dankreek:
**Extern:** Generated.


Deraen:
Do you have idea why extern changes are +1000/-5000? Api changes? Implementation changes that affect extern generation but the removed methods don't matter?

dankreek:
Hmm, that is alarming, isn't it? I'll try an figure out what happened here.

dankreek:
Did some checking on the diff and compared it with the Pixi 4.4.3 / 4.5.2 source. Looks like there was a lot of internal rejiggering between the 4.4 and 4.5 releases, most significantly, there was a `/lib` dir in 4.4.x that got rolled into the `/src` dir in 4.5.0. 

I'm in the middle of a project that relies heavily on PixiJS and contains 185 assertions, when I use my local copy of the Pixi 4.5.2 artifact that this PR generates, all my tests still pass.

PR #1206, opened 2017-06-13
add react-debounce-input

darkleaf:
New package:

**Extern:** Generated.


react-debounce-input/build.boot
```diff
@@ -0,0 +1,30 @@
+(def +lib-version+ "3.0.0")
+(def +version+ (str +lib-version+ "-0"))
+
+(set-env!
+ :resource-paths #{"resources"}
+ :dependencies   [['cljsjs/boot-cljsjs "0.5.2" :scope "test"]])
```
Deraen: In addition to deps-cljs react require, this package should also have a dependency on `cljsjs/react`.

darkleaf: Why?

My scenario.
I use previous version of react in my app.
I added cljsjs/react dep in my first commit in this PR.
But my app began to use new version form cljsjs/react-debounce-input.
And I wrote `[cljsjs/react-debounce-input "3.0.0-0" :exclusions [cljsjs/react]]`.

react-autosuggest [has no dependency from react](https://github.com/cljsjs/packages/blob/master/react-autosuggest/build.boot#L3).

Deraen: To ensure React is present, in a rare case where there is no other dependency on it. If you need specific version you should add exclusions or direct dependency.

https://github.com/cljsjs/packages/wiki/Foreign-library-dependencies

There are many bad packages as I don't always notice all the problems.

darkleaf: @Deraen I revert my commit.

Issue #1203, opened 2017-06-09
[material-components] Issues with new 0.12.1 update


ClashTheBunny:
Hey @r0man, 

I'm having trouble with 0.12.1.  I'm guessing it has to do with:
https://github.com/material-components/material-components-web/commit/e21ec90
and possibly with:
https://github.com/material-components/material-components-web/issues/767

Do the externs need to be updated for mdc-base?  Do I need to refer to mdc-base now instead of mdc?  What is the way it should be used?

This is my traceback when I change from 0.10.0 to 0.12.1:
```javascript
material-components.inc.js:90 Uncaught Error: Cannot find module "@material/base"
    at Object.<anonymous> (material-components.inc.js:90)
    at __webpack_require__ (material-components.inc.js:35)
    at Object.defineProperty.value (material-components.inc.js:81)
    at material-components.inc.js:84
    at webpackUniversalModuleDefinition (material-components.inc.js:14)
    at material-components.inc.js:15
(anonymous)	@	material-components.inc.js:90
__webpack_require__	@	material-components.inc.js:35
Object.defineProperty.value	@	material-components.inc.js:81
(anonymous)	@	material-components.inc.js:84
webpackUniversalModuleDefinition	@	material-components.inc.js:14
(anonymous)	@	material-components.inc.js:15
```

and

```javascript
Uncaught ReferenceError: mdc is not defined
    at Object.<anonymous> (material-components.cljs:79)
    at Object.reagent$impl$component$custom_wrapper_$_componentDidMount [as componentDidMount] (component.cljs:188)
    at react-dom.inc.js:4704
    at measureLifeCyclePerf (react-dom.inc.js:4514)
    at react-dom.inc.js:4703
    at CallbackQueue.notifyAll (react-dom.inc.js:948)
    at ReactReconcileTransaction.close (react-dom.inc.js:11361)
    at ReactReconcileTransaction.closeAll (react-dom.inc.js:14768)
    at ReactReconcileTransaction.perform (react-dom.inc.js:14715)
    at ReactUpdatesFlushTransaction.perform (react-dom.inc.js:14702)
(anonymous) @ material-components.cljs:79
mdc-textfield_componentDidMount @ component.cljs:188
(anonymous) @ react-dom.inc.js:4704
measureLifeCyclePerf @ react-dom.inc.js:4514
(anonymous) @ react-dom.inc.js:4703
notifyAll @ react-dom.inc.js:948
close @ react-dom.inc.js:11361
[...]

PR #1208, opened 2017-06-14
Update Material-components to 0.13.0, change to dosh to fail on failure

ClashTheBunny:
Update:

I have updated to 0.13.0, as that was released yesterday, and I also changed sh to dosh.

dosh allows circleci to fail if the build doesn't work.  It built fine on node 8.1.0 on my local MacMini.

This may solve https://github.com/cljsjs/packages/issues/1203.

### DIFF
--- a/material-components/build.boot
+++ b/material-components/build.boot
@@ -6,7 +6,7 @@
          '[boot.core :as boot]
          '[boot.tmpdir :as tmpd]
          '[clojure.java.io :as io]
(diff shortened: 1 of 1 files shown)


ClashTheBunny:
Both variants fail.  Is there a general problem with CircleCI accessing npm registry right now?

Deraen:
I updated CI setup Node yesterday, it is possible that broke something.

PR #1210, opened 2017-06-15
[slate] Update Slate to 0.20.3

bonkydog:


bonkydog:
Hmm, this PR has problems.  Sorry, will re-submit when I get them worked out.

PR #1216, opened 2017-06-18
[highlight] Update highlight.js to 9.12.0

pdlug:
Update:

Upgraded highlight.js to 9.12.0 which contains bug fixes and support for additional languages.


pdlug:
Like #1208 I'm also getting a CI failure, it works locally with node 7.9.x, 8.0.x, and 8.1.x.

PR #1217, opened 2017-06-19
Added highcharts-more extension dependency for highcharts

bjconlan:
New package:

**Highcharts-more:** Provided by the library.

Adds extra functionality to the existing highcharts library.

### DIFF
--- a/highcharts-more/README.md
+++ b/highcharts-more/README.md
@@ -1,0 +1,20 @@
+# cljsjs/highcharts-more
+
+[](dependency)
+```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
Please add dependency on `cljsjs/highcharts` and deps-cljs requires entry for `cljsjs.highcharts`.

Deraen:
Looking at this more closely, I think this should be provided as foreign-lib in the cljsjs/highcharts package. As should other files, like highcharts-3d.js and the modules.

bjconlan:
I'll leave this as a task for the highcharts maintainer.

PR #1197, opened 2017-06-07
[react-sticky] New package task, upgrade to 6.0.1, update externs

kennethkalmer:
* Based the new build task on the work done in react-truncate
* Tweaked webpack config to exclude ReactDOM & PropTypes
* Updated the README with a clear example of the new Sticky API
* Updated the README with extensive details on how to generate the externs

PR #1220, opened 2017-06-19
[semantic-ui-react] bump to 0.68.5

kennethkalmer:
Left everything else untouched


kennethkalmer:
I'm completely baffled why this project works without proper externs... I was updating the README with instructions similar to what I did for #1197 on how to update the externs when I noticed the diff was huge and the original externs is basically just the top level object.

PR #1270, opened 2017-07-28
Fix Skel v3

flyboarder:


Deraen:
Should the deps.cljs also include foreign-lib entries for skel-layout and skel-viewport?

flyboarder:
Yes! Didn't have the correct branch files in the PR!

PR #1288, opened 2017-08-15
[material-components] update version 0.17.0

ClashTheBunny:
Update:

Bump version to 0.17.0.  Builds fine on macos.


Deraen:
I think Circle is down currently. Will check again later.

PR #1300, opened 2017-08-23
[material-ui] Update to version 0.19.0

madvas:
MaterialUI doesn't need to include own React anymore


Deraen:
This is great!

PR #1302, opened 2017-08-23
[material-ui-chip-input] Update to version 0.17.0

madvas:

### DIFF
--- a/material-ui-chip-input/build.boot
+++ b/material-ui-chip-input/build.boot
@@ -1,9 +1,10 @@
 (set-env!
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]
-                  [cljsjs/material-ui "0.18.0-0"]])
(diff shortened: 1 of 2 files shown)


material-ui-chip-input/build.boot
```diff
@@ -61,6 +62,6 @@
      (minify :in  "cljsjs/material-ui-chip-input/development/material-ui-chip-input.inc.js"
              :out "cljsjs/material-ui-chip-input/production/material-ui-chip-input.min.inc.js")
 
-     (deps-cljs :name "cljsjs.material-ui-chip-input" :requires ["cljsjs.react" "cljsjs.material-ui"])
+     (deps-cljs :name "cljsjs.material-ui-chip-input" :requires ["react" "cljsjs.material-ui"])
```
Deraen: When using requires without `cljsjs.` prefix, the name should also skip the prefix.

madvas: yep, okay, changed

PR #1303, opened 2017-08-23
[react-infinite] Update to version 0.12.1

madvas:


react-infinite/build.boot
```diff
@@ -30,7 +31,7 @@
    (sift :include #{#"^cljsjs"})
 
    (deps-cljs :name "cljsjs.react-infinite"
```
Deraen: When using requires without `cljsjs.` prefix, the name should also skip the prefix.

madvas: yep, okay, changed

PR #1313, opened 2017-09-01
Updated react-leaflet to latest version 1.6.5

mharju:
Update:


Deraen:
The browserified files are available from unpkg, so no need to build these using webpack: https://unpkg.com/react-leaflet@1.6.5/dist/

mharju:
Great – I modified it to fetch those two files from the given location instead of a manual build.

react-leaflet/build.boot
```diff
@@ -1,14 +1,14 @@
...
+                  [cljsjs/react "15.6.1-2"]
+                  [cljsjs/react-dom "15.6.1-2"]
                   [cljsjs/leaflet "1.1.0-2"]])
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "1.4.1")
-(def +version+ (str +lib-version+ "-2"))
+(def +lib-version+ "1.6.5")
+(def +version+ +lib-version+)
```
Deraen: +version+ must include the build identifier.

mharju: OK! I added that now. I'm assuming it is for the current `+lib-version+`, so `-1` in this case?

PR #1319, opened 2017-09-06
update react-modal

hlolli:
Bump to version 2.3.1

**Extern:** Generated.
Update:

Api canged, now one needs to provide property for the RactModal calls `(.-default js/ReactModal)` for example. This is explained on the readme on their git repo ex. `var Modal = (require 'react-modal').default`


Deraen:
Where is use of ES6 default export documented? Doesn't make any sense, but I guess that is normal for JS libs.

hlolli:
No, bullshit from my behalf, this is not in the README, sorry, blind faith.

I should add this info into the cljsjs.react-modal readme?

Deraen:
I'm quite sure their webpack build is broken. ES6 export default property is not supposed to be seen on the dist build.

hlolli:
Ok, could be. This is where I actually found out about the .-default https://github.com/reactjs/react-modal/issues/470 and this is 19 days old, the version Im PR-ing against is 17 hours old. So between at least 2 versions, they haven't fixed it. The problem with the current cljsjs version of react-modal is that it uses deprecated `React.DOM.noscript` and fails. I'll compile the the new version in advanced and confirm that it works or not. Writing .-default may be better of 2 bad options.

hlolli:
Nah, looks like it's failing
```
TypeError: Cannot read property 'Component' of undefined
    at Object.defineProperty.value (eval at <anonymous> (:2:398), <anonymous>:97:19286)
    at t (eval at <anonymous> (:2:398), <anonymous>:97:444)
    at Object.t.exports (eval at <anonymous> (:2:398), <anonymous>:97:2088)
    at t (eval at <anonymous> (:2:398), <anonymous>:97:444)
    at Object.defineProperty.value (eval at <anonymous> (:2:398), <anonymous>:97:674)
    at t (eval at <anonymous> (:2:398), <anonymous>:97:444)
    at Object.defineProperty.value (eval at <anonymous> (:2:398), <anonymous>:97:532)
    at eval (eval at <anonymous> (:2:398), <anonymous>:97:537)
    at n (eval at <anonymous> (:2:398), <anonymous>:97:290)
    at Window.eval (eval at <anonymous> (:2:398), <anonymous>:97:312)
```
Hm, looks broken, twice they feed `_react.Component` to their function in their code, this could be the source (otherwise it's some other lib in my project causing this).

Deraen:
https://github.com/reactjs/react-modal/pull/493

hlolli:
Ok, it's not failing, version 2.3.1 works. The reason I got this message is unknown to me, but I'm building a Greasemonkey user script, and there it failed, in a script tag on html page, it works.

@Deraen nice hehe, even better :) I'll watch this thread and react if they fix this and update my PR.

hlolli:
PR Updated to v2.3.2, regenerated the externes and squashed the two commits.

PR #1322, opened 2017-09-07
update react-tooltip

hlolli:
* externs regenerated
* no api-change


Deraen:
Btw. you might be interested in using global-exports with these React libs, transition-group package has example for this: https://github.com/cljsjs/packages/blob/aa2ac970c50abbd0744a303bd8fa2adb61bd1715/react-transition-group_1/build.boot#L28-L34
https://github.com/cljsjs/packages/wiki/Unified-module-names-and-global-exports

hlolli:
Interesting

PR #1323, opened 2017-09-07
Add new package url-template (v2.0.8)

arbscht:
New package:

**Extern:** Generated.

### DIFF
--- a/url-template/build.boot
+++ b/url-template/build.boot
@@ -1,0 +1,30 @@
+(set-env!
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.7.0"  :scope "test"]])
+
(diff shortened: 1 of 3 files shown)


url-template/build.boot
```diff
@@ -0,0 +1,29 @@
+(set-env!
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.7.0"  :scope "test"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +version+ "2.0.8")
```
Deraen: You need to separate lib and build-identifier to two vars: https://github.com/cljsjs/packages/wiki/Creating-Packages#51-buildboot

arbscht: Oops, fixed.

PR #1333, opened 2017-09-22
[phoenix-html] Add Phoenix's HTML helper library

mudphone:
Extern file is empty on purpose.

**Extern:** Written by hand.


mudphone:
I'll squash and resubmit.

PR #1334, opened 2017-09-22
[phoenix-html] Add Phoenix's HTML helper library

mudphone:
Extern file is empty on purpose.

**Extern:** Written by hand.


Deraen:
> Extern file is empty on purpose.

So what is the purpose?

You can also just not have the extern file at all.

mudphone:
Oh, I didn't realize that. Thank you, I'll remove it and update the branch.

PR #1337, opened 2017-09-22
React-autosuggest updated  to v9.3.2

zarsky-broad:
Commit and PR based on https://github.com/cljsjs/packages/pull/1085.

Minimal api change, https://github.com/moroshko/react-autosuggest/releases/tag/v9.3.2

### DIFF
--- a/react-autosuggest/README.md
+++ b/react-autosuggest/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-autosuggest "9.0.0-0"] ;; latest release
+[cljsjs/react-autosuggest "9.3.2-0"] ;; latest release
 ```
 [](/dependency)
 
--- a/react-autosuggest/build.boot
+++ b/react-autosuggest/build.boot
@@ -9,9 +9,9 @@
          '[boot.util :refer [sh]]
          '[clojure.string :as str])
 
-(def +lib-version+ "9.0.0")
+(def +lib-version+ "9.3.2")
 (def +version+ (str +lib-version+ "-0"))
-(def +expected-checksum+ "E014B6B59C46C415F06A4277002541D6")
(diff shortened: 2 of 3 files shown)


zarsky-broad:
I can't figure out why this is failing on circle. It works locally.

Deraen:
Which node/npm are you running locally? Try with Node 8 / NPM 5.

```
npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.
npm WARN lifecycle react-autosuggest@9.3.2~prepublish: cannot run in wd %s %s (wd=%s) react-autosuggest@9.3.2 npm run dist && npm run standalone /root/.boot/cache/tmp/root/cljsjs/react-autosuggest/4l7/-i61w3b/react-autosuggest-9.3.2
```

zarsky-broad:
Interesting point, @Deraen. So, I tried running manually in the Docker image that Circle uses with the current state of `master`, and that same error occurs when running `boot package`. FWIW, I'm running the Node LTS, which is currently 6, locally, but it looks like that may actually be a permission issue…?

zarsky-broad:
@Deraen I figured it out! Since the user running `boot package` is `root`, `npm install` needs to be granted explicit permission to run scripts as `root` so they can access the temp directory boot creates. `--unsafe-perm` does that.

Deraen:
Cool. I also added global env variable to set this option, so packages don't need to care about this: https://github.com/cljsjs/packages/commit/25fb3bc2efaab88892f897cb34aa91b4174ebc06

zarsky-broad:
Fantastic. Thanks very much!

PR #1339, opened 2017-09-23
Update A-Frame to 0.7.0, fix description

AndresCuervo:

Also, used a search and replace to change 50 instances of `<x> - <y>` (not valid Javascript) into `['<x>.<y>']`.


AndresCuervo:
Actually, this PR isn't ready, noticed I need to make some manual alterations to the new externs. Will update when ready!

PR #1346, opened 2017-09-28
Add firebaseui

bbirec:
New package:
**Extern:** Generated.


firebaseui/resources/deps.cljs
```diff
@@ -0,0 +1,3 @@
+{:foreign-libs [{:file "cljsjs/development/firebaseui.inc.js"
```
Deraen: Instead of  creating `deps.cljs` by hand, it is preferred to use `deps-cljs` task from boot-cljsjs. It will make  set up the paths automatically (and ensure they are correct).

`(deps-cljs :name "cljsjs.firebaseui")`

PR #1347, opened 2017-09-28
[firebase] Updated Firebase to 4.4.0.

pbzdyl:
**Extern:** Included in Firebase distribution.


Deraen:
Seems fine, though looks like the new boot-cljsjs validate-checksums doesn't work with this package as it uses too string regex to select JS files.

pbzdyl:
Thank you!

PR #1356, opened 2017-10-03
Adds airbnb/enzyme

agzam:
This is my very first attempt to create a cljsjs package (please criticize - so I can learn). I desperately need to make it possible to use Enzyme in browser environment.

I need help with improving this package:

a) it currently only works with React 15
b) I was trying to find a way to use Global exports feature https://clojurescript.org/news/2017-07-30-global-exports ,so it could be required like this:

    (require '[cljsjs.enzyme :as enzyme])

but couldn't figure it out

### DIFF
--- a/enzyme/README.md
+++ b/enzyme/README.md
@@ -4,7 +4,7 @@ https://github.com/airbnb/enzyme
 
 [](dependency)
 ```clojure
-[cljsjs/enzyme "2.9.1"] ;; latest release
+[cljsjs/enzyme "3.0.0"] ;; latest release
 ```
 [](/dependency)
 
--- a/enzyme/README.md
+++ b/enzyme/README.md
@@ -1,0 +1,18 @@
+# cljsjs/enzyme
+
+https://github.com/airbnb/enzyme
+
+[](dependency)
+```clojure
+[cljsjs/enzyme "2.9.1"] ;; latest release
+```
+[](/dependency)
+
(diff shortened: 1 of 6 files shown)


Deraen:
Please go back to `deps-cljs`. If you want to use global-exports, update boot-cljsjs to 0.8.1.

agzam:
@Deraen Can you elaborate? You mean I should not be using deps.cljs? Why's that?

Deraen:
`deps-cljs` task is less lines and one less file, and ensures that file paths etc. are correct.

agzam:
@Deraen
This is ready to be merged. 
I hope you've seen my comment in Slack https://clojurians.slack.com/team/U0G75ARHC
I got back where I started (no global-exports). I guess I'm gonna write a wrapper in my app (that would be using this package), something like this:

    (ns app.enzyme
      (:require [cljsjs.enzyme]
                [reagent.core :as reagent]
                [reagent.ratom :as ratom]))

    (defn- wrapper [enzyme-fn form]
      (let [element (reagent/as-element form)
            node (enzyme-fn element)]
        {:root node
        :node node}))

    (defn shallow [form]
      (wrapper js/enzyme.shallow form))

    (defn mount [form]
      (wrapper js/enzyme.mount form))

    (defn simulate! [event-type {:keys [node root] :as wrapper}]
      (.simulate node event-type)
      (ratom/flush!)
      (reagent/flush)
      (.update root))

PR #1360, opened 2017-10-06
Added a small extension for leaflet which allows for rotated markers

Rovanion:
This is my first time trying to contribute a cljsjs library so feel free correct anything and everything, not that you should feel any less free had I contributed anything before.

New package:

**Extern:** Written by hand.

### DIFF
--- a/leaflet-rotatedmarker/resources/cljsjs/leaflet-rotatedmarker/common/leaflet-rotatedmarker.ext.js
+++ b/leaflet-rotatedmarker/resources/cljsjs/leaflet-rotatedmarker/common/leaflet-rotatedmarker.ext.js
@@ -1,0 +1,5 @@
+var rotatedMarker {
+	rotationAngle:     {},
+	setRotationAngle:  function () {},
+	setRotationOrigin: function () {}
+}
--- a/leaflet-rotatedmarker/resources/cljsjs/leaflet-rotatedmarker/common/leaflet-rotatedmarker.ext.js
+++ b/leaflet-rotatedmarker/resources/cljsjs/leaflet-rotatedmarker/common/leaflet-rotatedmarker.ext.js
@@ -1,5 +1,5 @@
-var rotatedMarker {
+var rotatedMarker = {
 	rotationAngle:     {},
 	setRotationAngle:  function () {},
 	setRotationOrigin: function () {}
-}
+};
--- a/leaflet-rotatedmarker/README.md
+++ b/leaflet-rotatedmarker/README.md
@@ -1,0 +1,18 @@
+# cljsjs/leaflet-rotatedmarkers
+
+[](dependency)
+```clojure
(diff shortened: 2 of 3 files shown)


leaflet-rotatedmarker/resources/cljsjs/leaflet-rotatedmarker/common/leaflet-rotatedmarker.ext.js
```diff
@@ -0,0 +1,5 @@
+var rotatedMarker {
```
Deraen:
```
./leaflet-rotatedmarker/resources/cljsjs/leaflet-rotatedmarker/common/leaflet-rotatedmarker.ext.js:1: ERROR - Parse error. Semi-colon expected
var rotatedMarker {
```

Deraen: Still broken, the error message is in fact confusing, the problem is missing `=`. Please test packages locally in future, with advanced compilation so Closure will try to read the extern.

Rovanion: Could you expand on the instructions on the Wiki by giving the complete command needed to do this local testing? I'm sure they are understandable by someone who's used boot before though. https://github.com/cljsjs/packages/wiki/Creating-Packages

Deraen: Complete commend to install the package to local Maven repository is mentioned there:

```
boot package install target
```

(target part is optional)

Then just add the dependency to project where you want to use the lib.

Rovanion: Sorry to bother you again but I've tested locally with `boot package install target` and the library seems to work when I use it in another project. It was the part about advanced compilation that I didn't get from the Wiki.

Deraen: The advanced part just means that you should use `:optimization :advanced` in the project where you use the library so that the extern gets used. As the extern file is still invalid JavaScript, if you don't see the problem currently, the extern file is not used.

Deraen: Anyway, as quick fix (if you are not able to test this with advanced compilation), change this line:

```
var rotatedMarker {
=>
var rotatedMarker = {
```

Rovanion: Oh, for some reason I thought I had already pushed that change. But it should be up now.

PR #1362, opened 2017-10-09
initial attempt to add howler to cljsjs

jborden:
I've found howler.js to be a simple sound library that "just works". I would like to have it included in cljsjs. 

New package:

**Extern:** Written by hand.

### DIFF
--- a/howler/build.boot
+++ b/howler/build.boot
@@ -18,10 +18,8 @@
 (deftask package []
   (comp
    (download :url (format "https://unpkg.com/howler@%s/dist/howler.js" +lib-version+)
-             :checksum "ac45047bc5f0b0b97aa22fd52bb29c5a"
              :target "cljsjs/howler/development/howler.inc.js")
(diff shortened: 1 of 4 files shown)


howler/build.boot
```diff
@@ -0,0 +1,29 @@
...
+       :version     +version+
+       :description "Javascript audio library for the modern web."
+       :url         "https://howlerjs.com/"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "https://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+   (download :url (format "https://unpkg.com/howler@%s/dist/howler.js" +lib-version+)
+             :checksum "ac45047bc5f0b0b97aa22fd52bb29c5a"
```
Deraen: No need to use `download :checksum` with `validate-checksums` task.

And there probably should be a `boot-cljsjs-checksums.edn` on the dir after running the package task, you should commit that to git.

Deraen:
Thanks

jborden:
Thank you for being super responsive!

PR #1369, opened 2017-10-17
[smoothscroll-polyfill] Update to 0.4.0 (latest).

briantrice:
Update:

### DIFF
--- a/smoothscroll-polyfill/README.md
+++ b/smoothscroll-polyfill/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/smoothscroll-polyfill "0.3.4-0"] ;; latest release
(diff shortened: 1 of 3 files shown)


Deraen:
Hmm, no need for extern file in this case, as code from this JS file is never called directly?

You can add `:no-externs true` option to `deps-cljs` to skip the check.

briantrice:
You can call the polyfill explicitly for some cases. I'm ambivalent about whether to expose it.

briantrice:
I should say that I want to force the polyfill in Chrome when its version of scrolling doesn't match what's requested. I may do that with the polyfill call since trying to set a flag before "require"ing it in Clojurescript is difficult.

Deraen:
Okay, I don't fully understand how this polyfill works, but if this works for you, this is fine.

PR #1382, opened 2017-10-30
Added missing Firebase externs

danielbraun:
The externs are taken from Firebase's externs in its repo.
Messaging specifically was missing, which is a main feature.

**Extern:** Included in Firebase distribution.


Deraen:

danielbraun:
OK, missed that part.

PR #1384, opened 2017-10-31
vega-embed 3.0.0-rc7

sorenmacbeth:
Update:


Deraen:
Btw, as note, it is not necessary to require all the three foreign-libs from your Cljs namespace, as vega-embed already depends on vega and vega-lite, those will be automatically used.

PR #1387, opened 2017-11-01
update to react-flip-move 2.9.17

hipitihop:
update to react-flip-move 2.9.17
- switched to using unpkg

Update: 


hipitihop:
@Deraen Many thanks for the prompt merge

PR #1388, opened 2017-11-01
[react-select] Update to 1.0.0-rc.10

svdm:
Caveat: this package does not include a build version in its version string. I left that alone for this version bump, but if desired I can add it.


Deraen:
There was [random failure](https://circleci.com/gh/cljsjs/packages/2692) on Clojars and the version didn't show up on Clojars web, though it was uploaded to the repo, I bumped the version to `-1` to be sure this was correctly deployed: https://clojars.org/repo/cljsjs/react-select/1.0.0-rc.10/

PR #1397, opened 2017-11-15
Emoji mart

bago2k4:
New package:

**Extern:** Generated.

EmojiMart is an emoji picker that supports different kind of emojis.


Deraen:
There seems to be some problems with browserify: 

```
+ browserify@14.5.0
added 175 packages in 15.518s
events.js:182
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open 'dist/emoji-mart.js.tmp-browserify-50017615376836266350'
sed: can't read dist/emoji-mart.js: No such file or directory
sed: can't read dist/emoji-mart.js: No such file or directory
```

bago2k4:
@Deraen i fixed the problem but the build is still failing for me on Ubuntu. It works on Mac.
It looks like it's something related to fsevents optional dependency.
Any thought?

Deraen:
Worked on CI now. Fsevents logs the warning on linux always, that isn't a problem.

PR #1404, opened 2017-11-17
[three-examples] Create package for threejs examples

sctincman:
New package: "example" scripts from threejs. These include useful things such as custom shader code and other filetype loaders (such as OBJ and GLTF loaders).

These scripts are not included in the threejs build process, and do not have minified versions. Minified versions are created at build time using [asset-minifier](https://github.com/yogthos/asset-minifier). Scripts are minified individually, and each script provides its own namespace. This was to allow developers to only include the pieces they need.

**Extern:** Generated. Package uses [externs-generator](https://github.com/jmmk/javascript-externs-generator) as part of the build process to create the single externs file.


sctincman:
> As the examples depend on Three.js, the package should have dependency on cljsjs/three and the foreign-libs should also require the cljsjs.three foreign-lib.

Done. Added to dependencies list and in requires for deps.cljs

> I think the best way is to separate it from package task and make sure it writes the new extern to resources folder, so git always contains the up-to-date file.

Ah ha, that is a good idea. Done.

> Boot-cljsjs provides minify task, no need to reimplement it here:

This was actually what I started with. The problem with that minify task is it only works on single input to single output. I needed to tweak it to allow the procedure to loop through the input fileset and minify each file separately. Is there a way to accomplish this in boot using a built-in task, something that loops a task for all input-files in fileset?  Further, there are a few ES6 files asset-minifier is unable to minify, and I needed a way to copy these when the minifier fails.

PR #1406, opened 2017-11-20
Add Microsoft Application Insights

hkjels:
New package:

**Extern:** Generated.

Update:


applicationinsights/build.boot
```diff
@@ -0,0 +1,28 @@
...
+      :scm         {:url "https://github.com/cljsjs/packages"}
+      :description "Microsoft Application Insights JavaScript SDK"
+      :url         "https://azure.microsoft.com/en-us/services/application-insights/"
+      :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+   (download :url (str "https://github.com/Microsoft/ApplicationInsights-JS/archive/master.zip")
+             :unzip true)
+   (sift :move {(re-pattern (str "ApplicationInsights-JS-master/dist/ai.js"))   "cljsjs/development/applicationinsights.inc.js"
```
Deraen: Best would be to add the library name to the path before development/production part:

`"cljsjs/applicationinsights/development/applicationinsights.inc.js"`

https://github.com/cljsjs/packages/wiki/Creating-Packages#6-writing-the-package-task

applicationinsights/build.boot
```diff
@@ -0,0 +1,28 @@
...
+                (re-pattern (str "ApplicationInsights-JS-master/dist/ai.0.js")) "cljsjs/production/applicationinsights.min.inc.js"})
```
Deraen: Is `0` the minified file of the latest version or what does it mean? The dist folder in zip seems to have all versions and in most cases there is `ai.version.js` and `ai.version.min.js` files.

applicationinsights/boot-cljsjs-checksums.edn
```diff
@@ -0,0 +1,2 @@
+{"cljsjs/production/applicationinsights.min.inc.js"
```
Deraen: Hmm, this file should include both files. Run `boot package target` and check that `cljsjs` folder really has the both files, if not, sift move patterns might not be moving the files correctly.

applicationinsights/build.boot
hkjels: Yeah. `0` is the minified version of the latest release. I'm not sure why it's like that

applicationinsights/boot-cljsjs-checksums.edn
hkjels: It works just fine on my box, but a colleague of mine is running into an issue on Windows with that sift-command he says

This is how it looks on my box:
```
.
├── common
│   └── applicationinsights.ext.js
├── development
│   ├── applicationinsights.inc.js
│   └── applicationinsights.inc.js.map
└── production
    ├── applicationinsights.min.inc.js
    └── applicationinsights.min.inc.js.map

3 directories, 5 files
```

I'll touch back when I know more about that error he's bumping into

hkjels: Even without the `unzip` and the `sift move` tasks it doesn't run properly on windows. But he's not able to package and install other libraries in `cljsjs` either, so I'm thinking it's more of a problem with his box.
Will this be any different when it's pulled in as a dependency?

PR #1408, opened 2017-11-21
Update D3.js and configure global exports

tow8ie:
Update D3.js to version `4.11.0`.

These changes include usage of the new `validate-checksum` function in conjunction with a `boot-cljsjs-checksums.edn` file. Also, the `:global-exports` option is now configured for D3.js. 

**Extern:** I updated the externs with generator and added some by hand.


tow8ie:
This PR needs [a fixed regex for the `validate-checksums` task][1] of `boot-cljsjs` to work. The update of the regex is not only relevant for the D3.js-package, but for all packages that will be updated in the future.

[1]: https://github.com/cljsjs/boot-cljsjs/pull/50

tow8ie:
D3.js released a new version just yesterday, so I also updated this package to D3.js version `4.12.0`.

PR #1423, opened 2017-12-10
fix(cljs/react-input-autosize): use requires with prefixes cljsjs

Atsman:
fixed the require items


Deraen:
New react and prop-types provide names without prefixes, they are correct.

PR #1424, opened 2017-12-11
[react] Add react-dom-test-utils

Peeja:
**Extern:** Not yet provided.

In React 16 the TestUtils addon was moved not to its own package but to ReactDOM, but built separately as `react-dom-test-utils.*.js`. This adds it as a fourth package within `react`.

I have a feeling there's a great way to generate the externs for this, but I don't know it. What's the best way to do it?

### DIFF
--- a/react/build.boot
+++ b/react/build.boot
@@ -7,9 +7,13 @@
 (def +lib-version+ "16.2.0")
 (def +version+ (str +lib-version+ "-0"))
 
-(def npm-project {'cljsjs/react "react"
-                  'cljsjs/react-dom "react-dom"
-                  'cljsjs/react-dom-server "react-dom"})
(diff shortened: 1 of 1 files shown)


Deraen:
Aha, previously test utils weren't packaged for browser (umd) but 16.2 added this. Hmm, I think we don't want additional packages, and I'd even like to remove react-dom-server package and move the server file to react-dom, to match npm packages.

Deraen:
See https://github.com/cljsjs/packages/pull/1425

I tested quickly the package with Reagent, but didn't try using test utils.

Deraen:
Sorry for not merging this, but I implemented this together with few other things myself.

Peeja:
That works, thanks!

PR #1428, opened 2017-12-12
Upgrade Slate to 0.31.3

EmergentBehavior:
Note that Slate was changed significantly from previous version on CLJSJS. Will be opening separate PRs to add `slate-react` and `slate-html-serializer`.

Update:

**Extern:** I updated the extern by hand.

PR #1429, opened 2017-12-12
Add slate-react

EmergentBehavior:
New package. Should probably not be merged until https://github.com/cljsjs/packages/pull/1428 is approved/merged since it has an explicit dependency on it.

New package:

**Extern:** Written by hand.

### DIFF
--- a/slate-react/build.boot
+++ b/slate-react/build.boot
@@ -3,7 +3,10 @@
 
 (set-env!
  :resource-paths #{"resources"}
- :dependencies '[[cljsjs/boot-cljsjs "0.8.2" :scope "test"]])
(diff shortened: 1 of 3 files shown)


slate-react/build.boot
```diff
@@ -0,0 +1,32 @@
+(def +lib-version+ "0.10.17")
+(def +version+ (str +lib-version+ "-0"))
+
+(set-env!
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.8.2" :scope "test"]])
```
Deraen: You should add dependency on `cljsjs/react`,  `cljsjs/react-dom` and `cljsjs/slate` here.

PR #1440, opened 2017-12-18
update lib version

kurt-o-sys:

New package:

**Extern:** Provided by the library.
**Extern:** Generated.
**Extern:** Written by hand.
**Extern:** Not needed as this is Closure library.

Update:

**Extern:** Not needed as this is Closure library.

### DIFF
--- a/react-bootstrap-table/build.boot
+++ b/react-bootstrap-table/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "4.3.0")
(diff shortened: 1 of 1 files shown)


kurt-o-sys:
ok, let's go for 4.2.0 ...

Deraen:
Please test locally: https://github.com/cljsjs/packages/wiki/Creating-Packages#7-install-your-package-locally-and-try-it-out

```
java.lang.IllegalStateException: Checksum of file v4.2.0.zip in not BA22769F79EC3E6A872D31FADFC64890 but 0623B30BAEF666ABD9D450218466FD9B
```

PR #1445, opened 2017-12-21
Upgrade ag-grid, ag-grid-react, ag-grid-enterprise to 15.0.0

benedekfazekas:
Update:

Also needed to manually specify clojure1.8 when building locally, may be some glitch in my local environment.


benedekfazekas:
:+1:

PR #1447, opened 2017-12-25
Update firebase to 4.8.1

korommatyi:
Update:


Deraen:
Please test locally and update checksums (old checksum file is probably empty as it was created with old boot-cljsjs):

```
Current checksums:
{}

New checksums:
{"cljsjs/development/firebase.inc.js"
 "1E7FCD644740FFC7F3EF0BC30E78920F"}
```

PR #1454, opened 2018-01-13
[accounting] Update externs

tessellator:
The existing externs for accounting would generate several warnings similar to the following when building a project with advanced optimizations enabled. This PR updates the externs to fix this issue.

```
WARNING - name accounting is not defined in the externs.
accounting.toFixed = function(value, precision) {}
^^^^^^^^^^
```

### DIFF
--- a/accounting/README.md
+++ b/accounting/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
Strange. Not sure why the previous externs wwould have given the warning, it should be okay to define object first and then extern the methods. Maybe Closure was confused as it was missing semicolons or something.

PR #1456, opened 2018-01-15
Update react-chartjs-2, chartjs to 2.7.0

lewang:
`boot package install` installed and tested in my specific usage.

globar JS var is renamed `reactChartjs2` -> `ReactChartjs2`

### DIFF
--- a/chartjs/build.boot
+++ b/chartjs/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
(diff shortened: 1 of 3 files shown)


lewang:
ok so this can't be combined, I'll open separate PRs.

lewang:
@Deraen I've followed the pattern of download the source and then building that I see for other libraries.

I'm wondering would it be acceptable to just install the library itself through NPM and rename files therein?  And the reasoning behingd why/why-not?

PR #1458, opened 2018-01-16
Update react chartjs 2 (redo of #1456)

lewang:
see #1456

### DIFF
--- a/react-chartjs-2/boot-cljsjs-checksums.edn
+++ b/react-chartjs-2/boot-cljsjs-checksums.edn
@@ -1,4 +1,4 @@
 {"cljsjs/react-chartjs-2/development/react-chartjs-2.inc.js"
- "984DDB67572FA50A6E8AAD286E082F4C",
+ "5B5D52232CB582A797AA42C225BA1A4D",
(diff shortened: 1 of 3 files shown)


lewang:
I'll ask this again here incase you missed it:

@Deraen I've followed the pattern of download the source and then building that I see for other libraries.

I'm wondering would it be acceptable to just install the library itself through NPM and rename files therein? And the reasoning behingd why/why-not?

Deraen:
So, the best thing for packaging this library will be to just use the existing UMD files, instead of building them here: https://unpkg.com/react-chartjs-2@2.7.0/dist/ (as documented on the wiki: https://github.com/cljsjs/packages/wiki/Creating-Packages#2-find-the-librarys-source-code)

If you would have to build package here, I don't think it matters much if you'd just install the package from npm or github. But the other packages probably download files from github, so it will be easiest to copy that. Also, the latest boot-cljsjs includes e.g. `run-commands` task will makes running `npm` easy: https://github.com/cljsjs/packages/blob/master/highlight/build.boot#L23

lewang:
@Deraen Ok, I've updated according to your recommendation.  Thanks for the explanation.

PR #1465, opened 2018-01-23
[Firebase] update to version 4.9.0 add firestore externs 

ericfode:
Just a routine update to add firestore support.

Also removed the node provider because it looks like the file it was using has been removed.


ericfode:
Puzzling over how to get firestore  exposed

ericfode:
Ok I think this is ready now.

PR #1472, opened 2018-02-01
Bump react-virtualized to 9.18.5

tonysherbondy:
- Noticed pulling react.dom instead of current react-dom. Seems like all cljsjs artifacts use react-dom now that cljsjs/react points to 16.

Paired with @sherbondy.


Deraen:
`cljsjs.react-dom` was error in React package, it has been fixed: https://github.com/cljsjs/packages/commit/41da295fd222ea1257a00ee5d6304c02976c5920#diff-266a1f7c2e2345169d3bc448da45eae6

tonysherbondy:
Gotcha. Updated.

PR #1478, opened 2018-02-07
[react-number-format] Update to 3.1.3

mulchy:


react-number-format/resources/cljsjs/react-number-format/common/react-number-format.ext.js
```diff
@@ -3,99 +3,103 @@
...
-        },
-        "value": {
-            "isRequired": function () {}
-        }
+  "defaultProps": {
+    "allowNegative": {},
+    "decimalSeparator": {},
+    "displayType": {},
+    "fixedDecimalScale": {},
+    "getInputRef": function () {},
```
mulchy: The diff looks worse than it is because the spacing changed, but the only real change in the externs was the addition of `getInputRef`

PR #1484, opened 2018-02-14
Updated quill to 1.3.5; added checksums

mkarp:
* Added `boot-cljsjs-checksums.edn` for quill
* Updated quill to 1.3.5
* Updated the description according to the official documentation
* Updated the API


mkarp:
@Deraen Thank you!

PR #1493, opened 2018-02-26
Upgrade material-ui to v1.0.0 beta.33 (material-ui@next)

craig-ludington:
* The `material-ui@next` package is a complete re-write of
  `material-ui`. Code that worked with version 0.19.2 will
  not work with `material-ui@next`.
* This version of `cljsjs/material-ui` does not provide
  react or react-dom; both were included in previous versions.


Deraen:
I think this should be packaged in separate directory, in case the 0.19 branch gets new updates. This can by creating copy of the package directory, and adding underscore and number to end of name, like `react_15`. This won't affect artifact name or anything, but makes possible to keep the packaging script for the other version.

craig-ludington:
Closing this pull request.  Please see https://github.com/cljsjs/packages/pull/1495 instead.

I made a separate branch and pull request to keep the commit history clean and readable.

PR #1495, opened 2018-02-27
Add material-ui_1 package (material_ui@next)

craig-ludington:
* Add the `material-ui@next` package as `material-ui_1`.
* Current version is `v1.0.0-beta.33`.
* The `material-ui@next` package is a complete re-write of
  `material-ui`. Code that worked with v0.19.2 will
  not work with `material-ui@next`.
* The original package, `material-ui` is unchanged; if you want
  v0.19.2 it's still available there.
* This version of `cljsjs/material-ui` does not provide
  react or react-dom; both were included in previous versions.


craig-ludington:
This replaces https://github.com/cljsjs/packages/pull/1493 following recommendations from @derean

PR #1496, opened 2018-02-27
[iota-mam] Add iotaledger/mam.client.js

transducer:
Adds [IOTA ledger's mam.client.js](https://github.com/iotaledger/mam.client.js). There is no official release of mam.client.js, so it uses the master.zip. Is this okay for now?

### DIFF
--- a/iota-mam/README.md
+++ b/iota-mam/README.md
@@ -1,0 +1,21 @@
+# cljsjs/iota
+
+[](dependency)
+```clojure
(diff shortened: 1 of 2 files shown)


Deraen:
I don't like such versions but I'll accept this for now. After I pressed merge I realized better way would be to use git SHA to download the zip and add that to version. We probably have few packages that work like that.

transducer:
@Deraen thanks. That is indeed a better way. I will keep it in mind when I will update the package.

PR #1489, opened 2018-02-17
[simple-peer] Added simple-peer@8.5.0

Nolan330:
New package:

**Extern:** Generated.

Simple WebRTC Peers. Tested locally.

PR #1504, opened 2018-03-09
Simple peer

qeshi:


Deraen:
More recent version was already added two weeks ago: https://github.com/cljsjs/packages/pull/1489

PR #1512, opened 2018-03-16
Update react-highlight to 1.0.7

danielcompton:
This isn't complete or ready to be merged. I've opened https://github.com/bvaughn/react-highlight.js/issues/11 to ask the original author to tag a new release, I've just forked the library and tagged it myself for testing.

I'm also not sure whether I've properly added PropTypes, I still get warnings in the console about them being used. This is my first CLJSJS package, does what I've done look correct?

### DIFF
--- a/react-highlight/build.boot
+++ b/react-highlight/build.boot
@@ -5,70 +5,28 @@
                   [cljsjs/prop-types "15.6.0-0"]
                   [cljsjs/highlight "9.6.0-0"]])
 
-(require '[cljsjs.boot-cljsjs.packaging :refer :all]
-         '[boot.core :as boot]
-         '[boot.tmpdir :as tmpd]
-         '[clojure.java.io :as io]
-         '[boot.util :refer [sh]])
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
 (def +lib-version+ "1.0.7")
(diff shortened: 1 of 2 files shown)


Deraen:
Can you paste the warnings about PropTypes? The dependency + requires look correct.

Btw. this package can be downloaded from NPM, no need for git tags or building the browserified JS here: https://unpkg.com/react-highlight.js@1.0.7/ E.g. prop-types build.boot provides simple example of what build script looks like when downloading files from unpkg.

And it is a bit unfortunate this package is called react-highlight here, because the npm package is react-highlight.js and there is another npm package called react-highlight.

Deraen:
@danielcompton Git tag seems to be available now, but check my previous comment about loading existing UMD module from unpkg.

danielcompton:
Thanks, I've simplified the build file which works nicely, but I'm still getting this error message in the console:

```
react-highlight.inc.js:162 Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
```

Deraen:
If such warning still happens, I don't see how it could be caused by Cljs or Cljsjs. You should test it on jsfiddle or something and report problems upstream if the warning also happens on other environments.

PR #1517, opened 2018-03-21
project0x/connect

fbielejec:
Packaging [project0x/connect](https://github.com/0xProject/0x-monorepo/tree/development/packages/connect)

Externs are generated as part of packaging task.

### DIFF
--- a/0xproject-connect/README.md
+++ b/0xproject-connect/README.md
@@ -1,0 +1,22 @@
+# cljsjs/zeroxproject-connect
+
+[](dependency)
+```clojure
+[cljsjs/zeroxproject-connect "0.6.6-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the ClojureScript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.zeroxproject-connect))
+
+(def http-client (new (aget js/connect "HttpClient") "https://api.com/")
+```
+
+Uses externs generated with by [`jmmk`'s Javascript Externs Generator'](https://github.com/jmmk/javascript-externs-generator) as part of the packaging process.
+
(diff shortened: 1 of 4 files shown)


connect/build.boot
```diff
@@ -0,0 +1,52 @@
...
+         '[boot.core :as boot]
+         '[boot.tmpdir :as tmpd]
+         '[clojure.java.io :as io]
+         '[boot.util :refer [sh]])
+
+(def +lib-version+ "0.6.6")
+(def +version+ (str +lib-version+ "-0"))
+
+(task-options!
+ pom {:project     'cljsjs/connect
```
Deraen: I think name should be `cljsjs/0xproject-connect` or something similar. We try to reflect npm names and npm Connect package is something completely different.

connect/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the ClojureScript compiler. After adding the above dependency to your project
+you can require the packaged library like so:
+
+Assumes following dependencies are present in the end users application:
+
+* [cljsjs/web3]()
```
Deraen: Don't assume. Add these as `:dependencies` to the package.

connect/build.boot
```diff
@@ -0,0 +1,52 @@
...
+      (binding [boot.util/*sh-dir* (str (io/file tmp))]
+        (do ((sh "npm" "install" "--include-dev"))
+            ((sh "npm" "run" "bundle"))
+            ((sh "npm" "run" "generate-extern"))
+            ((sh "rm" "-rf" "./node_modules"))))
+      (-> fileset (boot/add-resource tmp) boot/commit!))))
+
+(deftask package []
+  (comp
+   (build-connect)
```
Deraen: You could use `run-commands` task here to run those npm commands: https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L371 It will take care of setting up workding directory etc.

PR #1521, opened 2018-03-23
[elliptic] Added elliptic@6.4.0

Nolan330:


Nolan330:
yeah, apologies on that—wasnt sure if I should edit the generated extern. Is it exactly 2000-354663 that should be gone? I'm probably not knowledgeable enough to know what can be removed while still making the package useful

Nolan330:
upon looking into this further, the excessive line count comes from deeply nested parameters (points and constants) for each elliptic curve, that _each_ have _all_ of their respective functions and constants defined such as add, mul, neg, etc. If we got rid of these nested externs, would closure munge them such that
```
curve.deeply.nested.constant.mul(some.other.point)
```
might be broken? This is common when working with this library and shouldn't be risked. But I'm also am not a closure expert, so I could be mistaken and it may be safe to remove some of the nested levels.

Let me know your thoughts. Apologies again for the positively outrageous extern.

Deraen:
The objects which contain `closure_uid_785267531` properties are unnecessary at the least. Closure doesn't care about extern object nesting really,

```
curve.zeroBits = {}
curve.foo.bar.zeroBits = {}
```

Are the same. As long as zeroBits is mentioned somewhere in the file, it is not munged.

PR #1523, opened 2018-03-23
[bn] Added bn.js@4.11.8

Nolan330:


bn/build.boot
```diff
@@ -0,0 +1,27 @@
...
+  pom  {:project     'cljsjs/bn
+        :version     +version+
+        :description "BigNum in pure javascript"
+        :url         "https://github.com/indutny/bn.js"
+        :license     {"MIT" "https://github.com/indutny/bn.js#license"}
+        :scm         {:url "https://github.com/cljsjs/packages"}})
+
+(deftask package []
+  (comp
+    (download :url "https://wzrd.in/standalone/bn.js@4.11.8")
```
Deraen: Please use lib-version here so the url doesn't need to be updated separately:

`(download :url (format "https://wzrd.in/standalone/bn.js@%s" +lib-version+))`

PR #1524, opened 2018-03-23
[hash] Added hash.js@1.1.3

Nolan330:


hash/build.boot
```diff
@@ -0,0 +1,27 @@
...
+  pom  {:project     'cljsjs/hash
+        :version     +version+
+        :description "A few of many hashing algorithm implementations in plain javascript"
+        :url         "https://github.com/indutny/hash.js"
+        :license     {"MIT" "https://github.com/indutny/hash.js#license"}
+        :scm         {:url "https://github.com/cljsjs/packages"}})
+
+(deftask package []
+  (comp
+    (download :url "https://wzrd.in/standalone/hash.js@1.1.3")
```
Deraen: Please use lib-version here so the url doesn't need to be updated separately:

PR #1526, opened 2018-03-25
[buffer] update to use lib-version

Nolan330:


Deraen:
@Nolan330 https://wzrd.in is down for second day. Any idea what is going on? I'm going to need to remove all these pacakges if it doesn't start working soon. Or you could change the scripts to run webpack/browserify/whatever instead of depending on external service: https://github.com/cljsjs/packages/blob/3aa7b325004943fa9e481f905cf7ba8c5c4097f4/zeroxproject-connect/build.boot#L31

Deraen:
Looking here, seems like wzrd.in is very unreliable: https://github.com/browserify/wzrd.in/issues?q=is%3Aissue+is%3Aclosed

Nolan330:
Apologies on this. Didn't realize wzrd.in had a reputation for being unreliable. It had always worked (seemingly conveniently!) for me. As soon as I get a chance I'll convert the wzrd.in packages to browserify, which is all wzrd.in is anyway. Will be before the end of the week, if you can let them stick around until then. If not I can make new PRs no problem.

Deraen:
@Nolan330 I disabled the package for now to get CI running, remove these `disabled` files when you send the fixes: https://github.com/cljsjs/packages/commit/3f1b0420c8d58077ffb9e4e19cf72ea18e63e749

PR #1529, opened 2018-03-28
[stripe] Update stripe js to v3 and bump react-stripe-elements to v1.6

sherbondy:
Tentative PR!

I would appreciate some input/guidance on:

1. If the externs file for stripe.js that I generated seems legit / up-to-snuff since I used the unconventional route of passing a TypeScript Definitions file through tsickle to create it.

2. Whether it is better to split the stripe externs-only package into a separate stripe_3 package since it seem that stripe_2 isn't going away anytime soon (and could *hypothetically* get tweaked in a way that could impact the externs?), and stripe v3 is a radical departure...


Deraen:
Extern looks good. DefinitelyTyped + conversion should probably documented somewhere as it probably provides the best externs.

Yes, splitting packages seems like good idea.

sherbondy:
Ok, I've split out stripe_3 and reverted changes to the base stripe package so now `stripe_3` and `stripe` live alongside each other in the package listing. Both are referred to as `cljsjs.stripe` though.

Hope I did it correctly, thanks!

(react-stripe-elems still refers to stripe 2 package for now since stripe 3 is not on clojars.
I guess technically it'd be better to factor this out into 2 separate PRs? 
I can bump the minor [dash] version of react-stripe-elems to use stripe 3 once it's on clojars and submit another PR for it.)

PR #1533, opened 2018-03-28
[buffer] remove wzrd.in dependency

Nolan330:
I noticed I had duplicated the license field in the elliptic package.json, so that is bundled here to avoid a separate PR.


Deraen:
Oh this one was merged before wzrd.in went down, so no new package was built. If you want to build new version you should increment the build identifier: ` (def +version+ (str +lib-version+ "-1"))`

PR #1534, opened 2018-03-28
[hash] remove wzrd.in dependency

Nolan330:


Nolan330:
This should be the last of the wzrd.in dependencies. Apologies again (and future PR titles will be more consistent in their tense). Thank you for the quick workaround and maintaining cljsjs in general

Deraen:
You are welcome :)

Packages are now deployed. I updated docker image previously today and updated gpg on the image caused problems.

PR #1535, opened 2018-03-28
[mixpanel] Add cljsjs/mixpanel "2.19.0-0"

EmergentBehavior:
Adds CLJSJS version of MixPanel. Includes automatically generated externs and appropriate checksums.


EmergentBehavior:
I also included `people` manually in externs since normally MixPanel adds that dynamically.

PR #1545, opened 2018-04-02
[react-sanfona] Upgrade react-sanfona to 1.1.0-0

EmergentBehavior:
Just your friendly neighborhood upgrade.

Manually updated externs.

### DIFF
--- a/react-sanfona/build.boot
+++ b/react-sanfona/build.boot
@@ -2,8 +2,7 @@
   :resource-paths #{"resources"}
   :dependencies '[[cljsjs/boot-cljsjs "0.10.0"  :scope "test"]
                   [cljsjs/react       "16.3.0-0"]
-                  [cljsjs/react-dom   "16.3.0-0"]
-                  [cljsjs/react-dom-server "16.3.0-0"]])
+                  [cljsjs/react-dom   "16.3.0-0"]])
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all]
          '[boot.core :as boot]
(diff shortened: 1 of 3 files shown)


react-sanfona/build.boot
```diff
@@ -56,6 +58,8 @@
                  "cljsjs/react-sanfona/production/react-sanfona.min.inc.js"})
     (sift :include #{#"^cljsjs"})
     (deps-cljs :name "cljsjs.react-sanfona"
-               :requires ["cljsjs.react"])
+               :requires ["cljsjs.react"
+                          "cljsjs.react-dom"
+                          "cljsjs.react-dom-server"])
```
Deraen: Looking at React-Sanfona sources, it doesn't use react-dom/server. Foreign-lib requires aren't optimized so adding the react-dom-server here will make it part of output if react-sanfona is used, even if react-dom-server is not used.

react-sanfona/build.boot
```diff
@@ -1,15 +1,17 @@
 (set-env!
   :resource-paths #{"resources"}
-  :dependencies '[[cljsjs/boot-cljsjs "0.9.0"  :scope "test"]
-                  [cljsjs/react       "15.3.0-0"]])
+  :dependencies '[[cljsjs/boot-cljsjs "0.10.0"  :scope "test"]
+                  [cljsjs/react       "16.3.0-0"]
+                  [cljsjs/react-dom   "16.3.0-0"]
+                  [cljsjs/react-dom-server "16.3.0-0"]])
```
Deraen: Probably no need to require dom-server package here, it is deprecated and the react-dom/server file is not part of react-dom package.

PR #1549, opened 2018-04-03
[react-autosize-textarea] Upgrade react-autosize-textarea to 3.0.2

EmergentBehavior:
Upgraded to latest version.

Cleaned up `build.boot` to get validate checksums properly, use webpack-cli, adjust for source being in typescript.

@deraen Only question I have - The webpack process bundles react/react-dom into the built JS so let me know if we still need to explicitly indicate `cljsjs.react` and `cljsjs.react-dom`.

### DIFF
--- a/react-autosize-textarea/README.md
+++ b/react-autosize-textarea/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-autosize-textarea "0.4.8-0"] ;; latest release
+[cljsjs/react-autosize-textarea "3.0.2-0"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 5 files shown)


Deraen:
Webpack shouldn't bundle react/react-dom into the output as they are defined as externals in the webpack config. This is important because bundling them here would cause users to load React code several times when using Reagent or anything that uses React directly.

Deraen:
@EmergentBehavior Can you validate that the output bundle doesn't include React/-dom code? Based on the webpack configuration I'd way this is correct, but I'd like a confirmation :)

EmergentBehavior:
@Deraen Just double-checked. Should be all clear!

PR #1554, opened 2018-04-04
Upgrade tinycolor to 1.4.1, Moved to new boot-cljsjs

jbondeson:
Upgraded to 1.4.1. Also moved to latest boot-cljsjs and added checksums & deps.


tinycolor/resources/deps.cljs
```diff
@@ -0,0 +1,5 @@
+{:foreign-libs
```
Deraen: This file shouldn't exist when using `deps-cljs` task, which will generate the file.

jbondeson: Makes sense, I will update both PRs. Would you like me to rebase or will you squash on commit?

Deraen: Multiple commits are fine. I have stopped caring too much about best git commit/merge practices with this repo :)

jbondeson:
I went ahead and rebased to prevent the tragedy of the git log commons ;)

PR #1556, opened 2018-04-04
[react-onclickoutside] Upgrade to 6.7.1

EmergentBehavior:
* Generated checksums EDN
* Added required react-dom dependency
* Pulled proper development and production versions
* Rebuilt externs automatically


react-onclickoutside/build.boot
```diff
@@ -1,15 +1,15 @@
...
 (def +version+ (str +lib-version+ "-0"))
 
 (set-env!
-  :resource-paths #{"resources"}
-  :dependencies '[[cljsjs/boot-cljsjs "0.9.0" :scope "test"]
-                  [cljsjs/object-assign-shim "0.1.0-1"]
-                  [cljsjs/react "15.3.0-0"]])
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.10.0" :scope "test"]
+                 [cljsjs/object-assign-shim "0.1.0-1"]
```
Deraen: Why did you choose to add object-assign-shim dependency / require? Doesn't seem like mandatory requirement of react-onclickoutside. I think it is usually best to leave adding optional shims to the users, because if it is a dependency here, user can't remove the shim.

EmergentBehavior: @Deraen I didn't add it, it was already there and looks like it has been since this library was added to CLJSJS.

Deraen: Oh sorry, I misread the diff. Can you check if the dependency/require can be dropped?

EmergentBehavior: It looks like it was originally added to support React 15, but was later removed per https://github.com/Pomax/react-onclickoutside/commit/0f81bd40d0a7a39d6a71ca72667a8e61df075e5f. The only current usage I see is in a rollup configuration file https://github.com/Pomax/react-onclickoutside/blob/4a2684abfe84a6a88af28ea660118cb72e4609f4/rollup.config.js#L5. 

I can go ahead and remove it.

Deraen: The rollup file also just uses built-on Object.assign. The shim is for older browsers which don't have the built-in method.

PR #1558, opened 2018-04-04
Revert to previous format for building react-highlight package

danielcompton:
The previous commit in #1512 doesn't run properly, there is something about the new packaging method I used there that doesn't work correctly. I'd like to update to the new method, but in the meantime, here is a version for 1.0.7 which does work correctly, and uses the previous packaging method.


Deraen:
Looks like the official dist bundle uses CommonJS require to load externals: https://github.com/bvaughn/react-highlight.js/blob/master/webpack.config.dist.js#L14-L24

e.g. 
```js
/* 3 */
/***/ (function(module, exports) {

	module.exports = require("react");

/***/ }),
```

I think it should instead look for `window.React`.

PR #1563, opened 2018-04-06
[material-components] Update to 0.34.1

ClashTheBunny:
I believe there are large changes that happened in this version upgrade, including seeming to obviate the need for some of the externs data.

It also adds the checksum data, so this should be more secure than previously.

It works locally for what I've used older versions for.


Deraen:
Packaging the source-map file is unnecessary currently, but doesn't do any harm either. ClojureScript compiler isn't aware of foreign-lib source-maps so they aren't used.

Seems like you have one extra `)` on the end of the file.

ClashTheBunny:
I was trying to fix this with the inclusion of the source map:
image

Fixed the syntax error!  Sorry!

Deraen:
Interesting. Did including the map file fix this? I guess on this can work during development as the JS file is served as-is (not through Closure) and IF the map file somehow gets copied to somewhere browser can find it.

PR #1566, opened 2018-04-09
Datatables

tkauranen:


Deraen:
It is best if the packages use the same names as npm, e.g. datatables.net-dt, datatables.net-bs, datatables.net-buttons etc.

Button files should be all packaged in single datatables.net-buttons package, as that is how they are provided. Single package can provide multiple foreign libs using `deps-cljs` tasks `:foreign-libs` option like here: https://github.com/cljsjs/packages/commit/ed29223253b0dedf3f1d098cec0a60de65141588

If I understand datatables packaging correctly, there are 6 versions of the core datatables fire, for different styling frameworks: e.g dt variant for built-in and bs for bootstrap 3. And only one of those files should be loaded at the time? This probably makes packagin this bit a special case. E.g. datatables-buttons probably can't depend on the core because user has to choose if dt or bs version is used.

datatables-buttons/build.boot
```diff
@@ -0,0 +1,35 @@
...
+              :target "cljsjs/datatables-buttons/production/datatables-buttons.min.inc.js")
+    (download :url (format "https://cdn.datatables.net/buttons/%s/js/dataTables.buttons.js" +lib-version+)
+              :target "cljsjs/datatables/development/datatables-buttons.inc.js")
+    (download :url (format "https://cdn.datatables.net/buttons/%s/css/buttons.dataTables.min.css" +lib-version+)
+              :target "cljsjs/datatables-buttons/production/datatables-buttons.min.css")
+    (download :url (format "https://cdn.datatables.net/buttons/%s/css/buttons.dataTables.css" +lib-version+)
+              :target "cljsjs/datatables-buttons/development/datatables-buttons.css")
+
+	(deps-cljs :name "cljsjs.datatables-buttons"
+               :requires ["cljsjs.datatables"])
```
Deraen: As mentioned. This is probably problematic as this will load the built-in styled core file, instead of e.g. bootstrap styled one?

datatables-responsive-bs/build.boot
```diff
@@ -0,0 +1,35 @@
...
+              :target "cljsjs/datatables/development/datatables-responsive-bs.inc.js")
+    (download :url (format "https://cdn.datatables.net/responsive/%s/css/responsive.bootstrap.css" +lib-version+)
+              :target "cljsjs/datatables/development/datatables-responsive-bs.css")
+    (download :url (format "https://cdn.datatables.net/responsive/%s/js/responsive.bootstrap.min.js" +lib-version+)
+              :target "cljsjs/datatables/production/datatables-responsive-bs.min.inc.js")
+    (download :url (format "https://cdn.datatables.net/responsive/%s/css/responsive.bootstrap.min.css" +lib-version+)
+              :target "cljsjs/datatables/production/datatables-responsive-bs.min.css")
+
+	(deps-cljs :name "cljsjs.datatables-responsive-bs"
+               :requires ["cljsjs.datatables"])
```
Deraen: Should depend on `cljsjs.datatables-responsive` instead of core file?

PR #1569, opened 2018-04-10
[react-popper] Upgrade to 0.10.1

EmergentBehavior:
Upgrades react-popper and relevant react dependencies. Uses unpkg for source. Validates checksums.


EmergentBehavior:
Need to add another dependency...

EmergentBehavior:
Will fail until https://github.com/cljsjs/packages/pull/1571 is merged

PR #1572, opened 2018-04-10
[react-sanfona] Upgrade to 1.2.0

EmergentBehavior:
Yet another react-dom -> react.dom fix.

Also bumped react libs to latest CLJSJS patch.

### DIFF
--- a/react-sanfona/README.md
+++ b/react-sanfona/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-sanfona "1.1.0-0"] ;; latest release
+[cljsjs/react-sanfona "1.1.0-1"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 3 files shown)


EmergentBehavior:
There's actually a new version too so I'll just upgrade ...

Deraen:
Huh, surprising how many packages started using the bad name before it was fixed. I did quick `git grep` and looks like all should fixed now. By the way, instead of `validate-checksums` it is usually good idea to use `validate` task instead, which also validated the foreign-lib/externs entries and the files (Closure-compiler is used to read the JS files to make sure they can be parsed): https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L347

Thanks for all the fixes.

EmergentBehavior:
Thanks for the tip! I'll use validate in the future.

PR #1573, opened 2018-04-10
[react-infinite] Upgrade react-infinite to 0.13.0

EmergentBehavior:
* Bump to React 16 (supports it now)
* Latest deps versions
* Validate checksums
* Proper cljsjs namespacing

### DIFF
--- a/react-infinite/build.boot
+++ b/react-infinite/build.boot
@@ -1,14 +1,14 @@
 (set-env!
-  :resource-paths #{"resources"}
-  :dependencies '[[cljsjs/boot-cljsjs "0.9.0"  :scope "test"]
-                  [cljsjs/react "15.6.1-1"]
-                  [cljsjs/react-dom-server "15.6.1-1"]
-                  [cljsjs/react-dom "15.6.1-1"]
-                  [cljsjs/prop-types "15.5.10-0"]])
+ :resource-paths #{"resources"}
+ :dependencies '[[cljsjs/boot-cljsjs "0.10.0"  :scope "test"]
+                 [cljsjs/react "16.3.0-1"]
+                 [cljsjs/react-dom-server "16.3.0-1"]
+                 [cljsjs/react-dom "16.3.0-1"]
+                 [cljsjs/prop-types "15.6.0-0"]])
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "0.12.1")
+(def +lib-version+ "0.13.0")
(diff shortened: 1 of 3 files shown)


react-infinite/build.boot
```diff
@@ -21,17 +21,17 @@
...
+   (sift :move {#"^react-infinite-(.*)/dist/react-infinite.js$"
+                "cljsjs/react-infinite/development/react-infinite.inc.js"
+                #"^react-infinite-(.*)/dist/react-infinite.min.js$"
+                "cljsjs/react-infinite/production/react-infinite.min.inc.js"})
    (sift :include #{#"^cljsjs"})
-
-   (deps-cljs :name "react-infinite"
-              :requires ["react" "react-dom" "prop-types"])
+   (deps-cljs :name "cljsjs.react-infinite"
+              :requires ["cljsjs.react"
```
Deraen: The names without `cljsjs.` prefix are in fact correct. These names are used to match npm packages, so the libraries can use either cljsjs foreign-lib or npm modules, without changing their code (if the foreign-lib also supports global-exports). For backwards compatibility e.g. React provides both `react` and `cljsjs.react`. Packages using global-exports should depend the "new" name (`react`), and others can choose either.

As this package used to provide `react-infinite` I think it is best to provide the same name in new version, but we can also provide the "old" style name, this should do:

```
(deps-cljs :foreign-libs [{:file #"react-infinite.inc.js"
                           :file-min #"react-infinite.min.inc.js"
                           :provides ["react-infinite" "cljsjs.react-infinite"
                           :requires ["react" "react-dom" "prop-types"]
    :externs [#"react-infinite.ext.js"])
```

react-infinite/build.boot
```diff
@@ -21,17 +21,17 @@
...
+                         "cljsjs.react.dom"
+                         "cljsjs.prop-types"])
    (pom)
-
-   (jar)))
+   (jar)
+   (validate-checksums)))
```
Deraen: Try using `validate` here for foreign lib and JS file validation, in addtion to checksums.

react-infinite/build.boot
Deraen: https://github.com/cljsjs/packages/wiki/Unified-module-names-and-global-exports

EmergentBehavior:
@Deraen Thanks for taking a look! Will make both changes and push shortly.

If you're wondering, I'm upgrading many dependencies for one of my company's applications. That's why I'm firing off so many PRs!

EmergentBehavior:
@Deraen should be good to go

PR #1574, opened 2018-04-11
[recharts] Patch upgrade and cleanup

EmergentBehavior:
Including latest recharts dependency `1.0.0-beta.10-2` caused my application to fail until I made the following changes:

* bumped to latest react/react-dom dependencies
* Cleaned up build.boot
* Required 'react-dom' which is known peer dependency
* Used `validate` instead of `validate-checksums`.

### DIFF
--- a/recharts/build.boot
+++ b/recharts/build.boot
@@ -1,39 +1,33 @@
 (set-env!
  :resource-paths #{"resources"}
  :dependencies '[[cljsjs/boot-cljsjs "0.10.0" :scope "test"]
-                 [cljsjs/react "16.2.0-3"]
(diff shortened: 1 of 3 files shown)


recharts/build.boot
```diff
@@ -1,39 +1,33 @@
 (set-env!
  :resource-paths #{"resources"}
  :dependencies '[[cljsjs/boot-cljsjs "0.10.0" :scope "test"]
-                 [cljsjs/react "16.2.0-3"]
-                 [cljsjs/react-dom "16.2.0-3"]
-                 [cljsjs/react-transition-group "2.2.1-1"]
```
EmergentBehavior: This is no longer a required peer dependency.

PR #1576, opened 2018-04-13
[datatables.net]

tkauranen:


Deraen:
Phew, looks okay to me now, this must be one of the most complex packages here now. Thanks for fixing the PR.

PR #1580, opened 2018-04-13
[react-transition-group] Upgrade to 2.3.0

EmergentBehavior:
* Upgraded from 2.2.1-1 to 2.3.0-0
* Upgraded react peer dependencies to 16.x since it is supported
* Regenerated externs
* Added validation step and checksums


react-transition-group/build.boot
```diff
@@ -1,13 +1,13 @@
 (set-env!
-  :resource-paths #{"resources"}
```
Deraen: Minor note about indentation here, most of your PRs change indentation of at least `set-env!` call. I prefer indenting macro forms by two spaces, similar to functions: https://github.com/bbatsov/clojure-style-guide/pull/153 Probably your editor is currently configured to automatically change indentation for these forms.

Use of either style here is okay, but keeping the existing indentation of the file keeps the diff cleanest which helps with reviews a bit.

EmergentBehavior: Yeah it's applying a style I use elsewhere -- I'll be more conscientious of it in future PRs.

PR #1582, opened 2018-04-15
[popperjs] Patch upgrade 

EmergentBehavior:
Mixing in PopperUtils caused issues so it was removed.


EmergentBehavior:
Potentially relevant -> https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/index.js#L113-L121

If it's still an issue, I could try setting something like https://github.com/cljsjs/packages/blob/master/highlight/build.boot#L28-L35

Deraen:
Yes, I think providing two foreign libs here would be correct choice.

Single extern file can define externs for all foreign lib files, as it doesn't matter that externs to non existing JS objects.

https://github.com/cljsjs/packages/blob/master/mobile-drag-drop/build.boot#L42-L48 is another example using new-style deps-cljs task, providing two files (bit simpler than highlight example which uses regex wildcard to match many files).

Deraen:
> Yes, I think providing two foreign lib names here would be correct choice.

At least if you think making the `PopperUtils` object available. If not I can merge this as is.

PR #1594, opened 2018-04-26
added proton particle engine package

yogthos:

### DIFF
--- a/proton/README.md
+++ b/proton/README.md
@@ -1,0 +1,22 @@
+# cljsjs/proton
+
+[](dependency)
+```clojure
+[cljsjs/proton "3.1.2-0"] ;; latest release
+```
+[](/dependency)
+
+This jar comes with `deps.cljs` as used by the [Foreign Libs][flibs] feature
+of the ClojureScript compiler. After adding the above dependency to your project
+you can require and use the packaged library like so:
+
+```clojure
+(ns application.core
+  (:require cljsjs.proton))
+
+(def emitter (js/Proton.Emitter.))
+```
+
+Externs were generated using https://github.com/jmmk/javascript-externs-generator
+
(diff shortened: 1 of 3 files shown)


Deraen:
It would be best to name this `proton-js` to match the npm package.

yogthos:
👍

Deraen:
Thanks @yogthos 

For future packaging needs you might be interested in checking out this updated page about using npm-style names: https://github.com/cljsjs/packages/wiki/Npm-style-package-names-and-global-exports

yogthos:
Thanks for the quick merge, and the npm style guide looks helpful.

Deraen:
@yogthos How did you generate the extern file here? It looks like the full source code. It causes loads of Closure warnings and looking the code, I'm quite sure it doesn't work as extern because it creates several objects dynamically, so Closure is not able to read those as externs.

yogthos:
yeah I just ended up using the source directly for it, but you're right regarding the dynamic objects. It worked in my use case, but I should've looked a bit deeper.

Deraen:
@yogthos I emptied the file for now, but I didn't update the version number so no new release is made. The file was causing me issues with CI as [the extern validation script](https://github.com/cljsjs/packages/blob/master/test-extern-files.sh) printed several thousand lines of warnings which made debugging CI problems hard. For next release, the file should be fixed, or if you don't need extern for your use case (extern inference or something), and it is hard to generate one for this lib, it is also OK to leave the extern empty and mention this in readme (this is not optimal but there are couple of such packages).

yogthos:
Sounds good, I'll take a look at adding a proper externs file when I get a chance. I can see how the warnings are a pain. I disabled the warnings when using the source as the extern, but obviously that's not going to help in a general case.

PR #1605, opened 2018-05-09
added new office-ui-fabric-core package

binarykitchen:
here you go, another new package for office-js but comes with css assets only. hope it's okay?


Deraen:
So this is CSS only?

If you want to package it here it is okay. Though I'd suggest taking a look at https://www.webjars.org/, they have more automated setup to package npm/github projects automatically (from UI) into jar files. Tools that support using files from classpath (like cljsjs) should also support using files from Webjars.

PR #1615, opened 2018-05-18
[react-player] Added react-player package

vheuken:
Tested locally and it works.


react-player/build.boot
```diff
@@ -0,0 +1,30 @@
...
+ pom  {:project     'cljsjs/react-player
+       :version     +version+
+       :description "A React component for playing a variety of URLs, including file paths, YouTube, Facebook, Twitch, SoundCloud, Streamable, Vimeo, Wistia and DailyMotion"
+       :url         "https://github.com/CookPete/react-player"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+    (download :url (str "https://unpkg.com/react-player@1.5.1/dist/ReactPlayer.js"))
```
Deraen: Please use versioned url to download the archive or the files. Use the `+lib-version+` var you have defined to build the url:

```clj
(def +lib-version+ "1.2.3")

...

          :unzip true)
```

vheuken: Fixed.

PR #1621, opened 2018-05-29
[firebase] General improvements/fixes to build.boot

gabrielsimoes:

### DIFF
--- a/firebase/README.md
+++ b/firebase/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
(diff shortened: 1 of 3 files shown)


Deraen:
@gabrielsimoes Note that version number variables must be written exactly as they where, because of brittle CI deploy scripts: https://github.com/cljsjs/packages/commit/c6af34afd2b03df216df583a4a18bc931423b170

Unfortunately CI tests don't catch there problems.

PR #1626, opened 2018-06-05
[react] Upgrade react to 16.4.0

EmergentBehavior:
I did not update the externs as it appears they are generated differently than other packages and the ones automatically generated using React via unpkg are unlike the ones we have currently (they are much shorter).

@Deraen Let me know what is needed to upgrade externs (if it's anything special) and I can update this PR.

I did notice it looks like we have 'React 15' externs at the top then additions for 'React 16.3' on the bottom. I'm unaware if that is an intentional separation.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react "16.3.2-0"] ;; latest release
+[cljsjs/react "16.4.0-0"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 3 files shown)


Deraen:
Extern are written by hand and I keep the up-to-date by following changelog: https://github.com/facebook/react/blob/master/CHANGELOG.md#1640-may-23-2018

This release shouldn't have any changes that change the API.

Could you fix this PR to keep the old indentation?

EmergentBehavior:
Will do

EmergentBehavior:
@Deraen Ready for another look

PR #1637, opened 2018-06-07
[tfjs] Add tfjs 0.11.5

namenu:
TensorFlow.js is A WebGL accelerated JavaScript library for training and deploying ML models.


Deraen:
This looks like extern-only package? Was this the intention?

namenu:
Yes, I intended since it's still under 1.0.0.
I wasn't sure if I should upload empty checksum file.

Deraen:
Umm, how does the version affect this? How is the JS for TensorFlow.js added to the browser?

namenu:
It is added by `<script>` tag with version number.
e.g.) https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@0.11.6
(In the meantime, anew version has been released)

I am thinking that it has not enough reasons to be extern-only, so I have added `download` process.

Please take a look !

Deraen:
Now that `deps-cljs` task is used, the `deps.cljs` file is unncessary. It will be rewritten by the task.

PR #1647, opened 2018-06-21
[@material-ui/icons] Adds in the @material-ui/icons package

jimberlage:
This just adds the material UI icons back in, now that the `@material-ui/core` package has been updated.


Deraen:
Thanks

PR #1650, opened 2018-06-27
create new package js-joda-locale-en-us

henryw374:

### DIFF
--- a/js-joda-locale-en-us/build.boot
+++ b/js-joda-locale-en-us/build.boot
@@ -1,0 +1,30 @@
+(set-env!
+  :resource-paths #{"resources"}
+  :dependencies '[[cljsjs/boot-cljsjs "0.5.2"  :scope "test"]])
+
+(require '[cljsjs.boot-cljsjs.packaging :refer :all])
+
+(def +lib-version+ "1.0.0")
+(def +version+ (str +lib-version+ "-0"))
+
+(task-options!
+ pom  {:project     'cljsjs/js-joda-locale-en-us
+       :version     +version+
+       :description "prebuilt en-US locale addon for js-joda"
+       :url         "https://js-joda.github.io/js-joda/js-joda-locale"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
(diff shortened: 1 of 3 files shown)


Deraen:
It would be best to package locales similar to npm, i.e. `js-joda-locale` is single package, which contains all the locales.

Highlight package should provide example on how to create packages providing several files: https://github.com/cljsjs/packages/blob/master/highlight/build.boot

henryw374:
Hi, thanks for picking this up. Although js-joda-locale is a single package, it depends on others and bundling the dependencies with all locales, the total package size would be huge (MBs). Because of that, js-joda-locale itself comes pre-packaged into locale-specific packages: https://github.com/js-joda/js-joda-locale#use-prebuilt-locale-packages.

This en-us package should cover 90%+ of use cases, (hopefully not being too English-centric) where you need to format dates with M for month 'dd MMM yyyy', whilst keeping the size down to a few KBs.

Deraen:
It doesn't matter if the package is huge, only the files required by application project are included in build output.

henryw374:
I should point out that these packages are not yet on npm ... the maintainer is having a few issues: https://github.com/js-joda/js-joda-locale/issues/78 . So i could just publish this outside of cljsjs for now if that's preferred

js-joda-locale-en-us/build.boot
```diff
@@ -0,0 +1,30 @@
...
+ pom  {:project     'cljsjs/js-joda-locale-en-us
+       :version     +version+
+       :description "prebuilt en-US locale addon for js-joda"
+       :url         "https://js-joda.github.io/js-joda/js-joda-locale"
+       :scm         {:url "https://github.com/cljsjs/packages"}
+       :license     {"MIT" "http://opensource.org/licenses/MIT"}})
+
+(deftask package []
+  (comp
+    (download :url (format "https://raw.githubusercontent.com/js-joda/js-joda-locale/8e384141cb191c8cbf35b2844cab15ace035f806/packages/en-us/dist/index.js" ))
```
Deraen: Whatever file is downloaded (Js or zip) the url should use version number, not sha.

henryw374: unfortunately this is not available yet. I'll close this PR and post another one when the en-us prebuilt is on npm.

Thanks

PR #1670, opened 2018-08-01
[antd] Update antd to version 3.7.3

priornix:
[antd] Update antd to version 3.7.3


miikka:
I just merged PR #1676, which upgrades antd to 3.8.0, so I'll close this one.

PR #1675, opened 2018-08-04
[tfjs] bump 0.12.0

namenu:


tfjs/build.boot
```diff
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "0.11.6")
+(def +lib-version+ "0.12.0")
 (def +version+ (str +lib-version+))
```
miikka: Could you change this so that it has the cljsjs version `-0` appended like the other packages? See the wiki here: https://github.com/cljsjs/packages/wiki/Creating-Packages#51-buildboot

PR #1682, opened 2018-08-08
[textures] Add d3 dependency and lock version of js file source.

ibothwell:
Added dependency on d3 to remove requirement for external include and better integration with advanced compile. Lock version of the source js file to ensure build reproducibility.

### DIFF
--- a/textures/README.md
+++ b/textures/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/textures "1.0.3-0"] ;; latest release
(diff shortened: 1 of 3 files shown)


ibothwell:
@miikka Thank you for the feedback. I took all the suggestions you presented and refactored the package step to make it cleaner by leveraging the minify task.

- Updated `+lib-version+` to `1.2.0`
- Locked download to archive version equal to `+lib-version+`
- Removed `:checksum` in favour of `(validate)`
- Added `(minify)` for generation of the production .js

PR #1689, opened 2018-08-09
[vega-tooltip] Update to 0.12.0 - most recent compatible with vega 3.3.1

jsa-aerial:
Update to release 0.12.0 - version used by  vega 3.3.1


vega-tooltip/build.boot
```diff
@@ -17,6 +17,29 @@
        :scm         {:url "https://github.com/cljsjs/packages"}})
 
 (deftask package []
+  (task-options! push {:ensure-branch nil})
+  (comp
+    (download
+     :url (format "https://unpkg.com/vega-tooltip@%s/build/vega-tooltip.js" +lib-version+)
+     :checksum "9DA40C6DFB095BECD7DED0C1140306C7")
```
miikka: By the way, instead of including these checksums here inline, you can add task `(validate)` as the final thing in the `package` task, after `(jar)`. Then Boot will automatically store the checksums in a separate file `boot-cljsjs-packages.edn` and ask you if you want to update when you're building the package with `boot package`. See my recent pull request #1680 for an example.

It's not necessary to change it, but it would make updating this package the next time a bit easier.

jsa-aerial:
OK, something isn't quite right with this.

First, that git command doesn't work. It fails with 'file not found' no matter where you try it. I had to manually create the new `vega-tooltip` subdir and manually move the `common` under it.

Next, just making the sifts go to the new `vega-tooltip` subdir does not actually work for the dev and prod paths. This:

```
    (sift :move {#".*vega-tooltip\.js$"      "cljsjs/vega-tooltip/development/vega-tooltip.inc.js"})
    (sift :move {#".*vega-tooltip\.min\.js$" "cljsjs/vega-tooltip/production/vega-tooltip.min.inc.js"})
    (sift :move {#".*vega-tooltip\.css$"     "cljsjs/vega-tooltip/common/vega-tooltip.css"})
```
results in:
```
     0 Fri Aug 10 10:37:14 EDT 2018 cljsjs/vega-tooltip/
     0 Fri Aug 10 10:37:14 EDT 2018 cljsjs/vega-tooltip/common/
 19828 Thu Aug 09 10:00:40 EDT 2018 cljsjs/vega-tooltip/common/vega-lite.ext.js
     0 Fri Aug 10 10:37:14 EDT 2018 cljsjs/development/
531164 Fri Aug 10 10:37:13 EDT 2018 cljsjs/development/vega-lite.inc.js
     0 Fri Aug 10 10:37:15 EDT 2018 cljsjs/production/
173167 Fri Aug 10 10:37:13 EDT 2018 cljsjs/production/vega-lite.min.inc.js
```

You can even _manually_ create `.../vega-tooltip/development` and `.../vega-tooltip/production` and the resulting jar still does not have the dev and prod under `vega-lite` subdir.

As I mentioned, I'm not a 'boot guy' so maybe there is some obvious boot related thing I am missing???

jsa-aerial:
Another potentially odd thing about this, is that even without that extra subdir, everything just works - deps, requires, advanced optimization, etc.

jsa-aerial:
Looking at this some more, I see the situation is worse than I noted above. I hadn't noticed before that with the new sifts, the css file just disappears altogether. So, the setup as currently suggested actually no longer works.

miikka:
I'm not sure what you mean by CSS disappearing?

If this is getting too complicated, let's just ignore this directory structure issue for now. We do not have to fix *everything* at once. As you noticed, the package will work for most users (like the previous versions did) although e.g. [ring-cljsjs](https://github.com/Deraen/ring-cljsjs) users may have problems.

If you just remove the commented-out code, and if the CSS works, I'd be happy to merge this .

jsa-aerial:
> I'm not sure what you mean by CSS disappearing?

Turns out this is what I refer to as 'chasing a ghost'. I was in the wrong directory when looking at the results and so I thought I saw something but it wasn't really there...

OK, I removed the commented code and now have the lib subdir working just fine as well. Will commit this new try. Left the old fashioned checksums for now.

PR #1693, opened 2018-08-17
Initial import of react-test-renderer-shallow.

mikesperber:
SSIA: Another package spun out of React; needed this for Reacl.


mikesperber:
Yay, thanks!

PR #1698, opened 2018-08-27
[iota-mam] fix externs name

transducer:
It conflicted with iota.js:

when building a project that uses both iota and iota-mam with advanced compilation the extern files conflicted. The reason was that iota-mam had an externs file `iota/iota.ext.js` and that is the same as the one for iota. This commit changes the iota-mam externs file to `iota-mam/iota-mam.ext.js`.

### DIFF
--- a/iota-mam/build.boot
+++ b/iota-mam/build.boot
@@ -5,7 +5,7 @@
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
 (def +lib-version+ "0.0.1")
-(def +version+ (str +lib-version+ "-0"))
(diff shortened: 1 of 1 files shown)


Deraen:
The branch seems to also include new leaflet-polylinedecorator package where extern seem to be incorrect.

transducer:
You are right. I will split them up in two separate pull requests.

transducer:
I removed the leaflet-polylinedecorator commits to https://github.com/cljsjs/packages/pull/1699.

Deraen:

transducer:
Done

PR #1699, opened 2018-08-28
[leaflet-polylinedecorator] Add leaflet-polylinedecorator

transducer:
I put externs in leaflet externs because I don't know how to extend L (if it is possible at all).

It does work for my use case (adding leaflet + polylinedecorator) and using it.

### DIFF
--- a/leaflet-polylinedecorator/README.md
+++ b/leaflet-polylinedecorator/README.md
@@ -1,0 +1,16 @@
+# cljsjs/iota
+
+[](dependency)
+```clojure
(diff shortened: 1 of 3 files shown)


transducer:
Ah maybe I can extend `L` using:

```
L.PosAnimation = function () {};
L.Symbol.marker = function() {};
L.polylineDecorator = function() {};
```

transducer:
I did that and changed the commit. It works.

Deraen:
You should also add leaflet dependency and deps.cljs require to leaflet: https://github.com/cljsjs/packages/blob/master/leaflet-locatecontrol/build.boot

transducer:
Done

PR #1705, opened 2018-09-11
Bump firebase version

danielstockton:

### DIFF
--- a/firebase/build.boot
+++ b/firebase/build.boot
@@ -4,7 +4,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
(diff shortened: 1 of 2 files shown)


firebase/build.boot
```diff
@@ -4,7 +4,8 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "5.0.4")
+(def +lib-version+ "5.4.2")
+'use strict';
```
Deraen: You have accidentally added a line of JS here.

danielstockton: Oops, thanks.

PR #1709, opened 2018-09-14
[react-json-view] Add minified version of js

ignorabilis:
[react-json-view] 
Add minified version of js


react-json-view/boot-cljsjs-checksums.edn
```diff
@@ -0,0 +1 @@
+{}
```
Deraen: This file is missing contents. Please run `boot package` and commit updated file.

ignorabilis: @Deraen - I ran `boot package` from Windows; while it worked I couldn't find a way to debug why checksums were left empty. Anyway, I generated these via a Linux Docker container so it should be ok now.

Deraen: Sorry about windows support. Probably this problem was caused by boot-cljsjs not taking Windows path delimiter into account in this regex which is used to select which files to add to checksum file: https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L281

I tried to fix this in new version but not sure if it works yet: https://github.com/cljsjs/boot-cljsjs/commit/e538c38992e2baeaf1d4c282edae63ff85a81899

ignorabilis: @Deraen - I just downloaded and tested, it's working now.

PR #1716, opened 2018-09-24
Move html-screen-capture -> html-screen-capture-js

miikka:
The packaging scripts assume that the directory name and the asset name match each other. Wiki says that `js` should be dropped from the name, but @Deraen says that the package name should match the NPM name. Let's go with the latter, then.

This should fix the Circle build.


Deraen:
Done already: https://github.com/cljsjs/packages/commit/9d082a1cecd30330659ae6069c9efb7df33c6d90

PR #1718, opened 2018-09-26
Bump material-ui

jimberlage:
This updates material-ui_1 and material-ui-icons (which goes with it) to the 3.x.x version.  The only breaking change was that a path to some of the icons had changed, which shouldn't affect people's CLJS code.  They skipped version 2 entirely, not sure why.


jimberlage:
Advanced compilation revealed that a lot of this library isn't specified in externs.  I'm gonna fix that real quick.

Deraen:
I tested the new version with Reagent Material-UI example and this seems to work. I remembered this package mentioned that the externs aren't complete, but I was wrong and readme doesn't mention this.

If you are able to extend the externs that is cool. But I can also accept this as is.

jimberlage:
I'll fix that build.boot as well.  I need advanced compilation so I'll just take on the work of completing the externs.

jimberlage:
That should be good.  I pulled the externs from

- https://github.com/mui-org/material-ui/blob/v3.1.1/packages/material-ui/src/index.js
- https://github.com/mui-org/material-ui/blob/v3.1.1/packages/material-ui/src/colors/index.js
- https://github.com/mui-org/material-ui/blob/v3.1.1/packages/material-ui/src/styles/index.js

PR #1733, opened 2018-10-10
Update dependency to pouchdb 7.0.0

maacl:


Deraen:
You should also increment `+lib-version+` so that new version is deployed to clojars:

```
```

https://github.com/cljsjs/packages/wiki/Updating-packages

Deraen:
Ah right, this was included in the other PR. I updated the build identifier myself so new build should be deployed to clojars soon.

PR #1738, opened 2018-10-22
[pikaday] Update to version 1.8.0

djtango:

### DIFF
--- a/pikaday/build.boot
+++ b/pikaday/build.boot
@@ -5,22 +5,22 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "1.5.1")
-(def +version+ (str +lib-version+ "-2"))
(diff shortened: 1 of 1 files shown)


pikaday/build.boot
```diff
@@ -5,22 +5,21 @@
...
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "1.5.1")
-(def +version+ (str +lib-version+ "-2"))
+(def +lib-version+ "1.8.0")
 
 (task-options!
   pom {:project 'cljsjs/pikaday
-       :version +version+
+       :version +lib-version+
```
Deraen: `+version+` var must not be removed. Just keep it like previous and reset build identifier (the part after dash) to zero.

djtango:
@Deraen 

Have made the requested changes to not remove `+version+`

PR #1505, opened 2018-03-12
[vega] Update vega related packages

metasoarous:
This commit specifies the following updates:

vega -> 3.2.1
vega-lite -> 2.2.0
vega-tooltip -> 0.6.1
vega-embed -> 3.1.1

These updates collectively resolve issues with rendering rule marks
appropriately, and generally put things on the latest releases.

Tested out with github.com/metasoarous/oz and appears to fix the issues I was having.

PR #1747, opened 2018-11-08
[vega] Update vega related packages

metasoarous:
Versions: vega -> 4.3.0, vega-lite -> 3.0.0-rc8, vega-embed -> 3.23.1, vega-tooltip -> 0.13.0


metasoarous:
Hello @Deraen 

@jsa-aerial [pointed out](https://github.com/metasoarous/oz/commit/f6d53944381cb0750bf3e539efb7154cc4bf56d4#commitcomment-31224677) that updating multiple projects at once like this may not work. However, I've previously submitted a single PR like this (#1505), which you merged and didn't seem to have a problem with. Was the success there an accident, or has something changed? I certainly don't want to be a bother, so please let me know if I should break this up into multiple PRs.

jsa-aerial:
Hmmmm - looks like this went through (where's the 'thinking face' emoji?)...

jsa-aerial:
OK, I know why this 'passed' - @metasoarous you forgot to update the dependencies! for vega-lite, vega-embed, and vega-tooltip!!

Also, I believe your new externs will fail if you actually try to compile them - the generator messes up the final comment box. Did you actually try this stuff in an advanced compiled cljs example? I'm pretty sure it will fail...

metasoarous:
@jsa-aerial Lol; Sure enough. I pushed an update with all the requirements properly set, and the tests are failing now as you suspected. I'll break this up and resubmit.

I'm not sure what you mean about the externs though. What is messed up with the comment box? I have run this through advanced compilation and it seems to be working.

metasoarous:
@Deraen OK; this PR has been modified to include just the Vega updates, and it looks like the checks are passing. Should have two more PRs in flight momentarily.

jsa-aerial:
The final comment box is structured as /* ..../*....*/ That does not work as it is NOT legal JavaScript. You need to get rid of the middle '/'.

Deraen:
I fixed the comment so hopefully this works now.

@metasoarous I occasionally merge PRs with failing tests if I know they are failing because they update several packages with dependencies in each other, and I can reasonably see there isn't other problem with them. But splitting such updates to multiple PRs makes review easier.

Deraen:
Btw @jsa-aerial @metasoarous adding `validate` task will run the extern JS through Closure which should catch any syntax errors: https://github.com/cljsjs/packages/commit/b15e2562b2e77ab62a02cb9436234844be795974

In this case the comment wouldn't have broken the test, as I think JS allows `/` in start of multi-line comments:
```
/*/*      -> SyntaxError

/*
/*/      -> Works
```

jsa-aerial:
@Deraen I've had that comment fail when running through Closure, which (since the jar was not built with `validate` enabled) only occurred upon building an artifact using that jar with `:optimizations :advanced` enabled. The only way I could fix it was by fixing that comment section.

PR #1749, opened 2018-11-09
[vega-lite] update vega-lite version to 3.0.0-rc8-0

metasoarous:
Compiled against Vega 4.3.0-0 (see #1747).


jsa-aerial:
@Deraen With respect to the PRs for vega-lite and vega-tooltip, can we (I suppose it is really you) fix the ext.js final block comment like you did for vega?

Also I have a new 3.24.1 vega-embed that I would like to commit and PR. So,  I don't know how that affects the current PR for version 3.23.1. If you think it makes sense to merge that one first, it would be good to fix the comment block in the ext.js for that one as well.

I have a  couple of libs awaiting release depending on these things

Thanks!

PR #1756, opened 2018-11-28
Init d3-sankey at 0.7.1

timds:
Also add notes to d3 repository about d3 plugin externs


timds:
Sorry I didn't realise the externs could be merged like that. I've fixed the commit now.

timds:
Sorry I didn't realise the externs would be merged like that, partially because I seemed to have some trouble with `:global-exports {d3-sankey d3}` technique, so I've reverted to the legacy method.

PR #1768, opened 2018-12-22
[recharts] Update to v1.4.2

rarous:

### DIFF
--- a/recharts/build.boot
+++ b/recharts/build.boot
@@ -7,8 +7,8 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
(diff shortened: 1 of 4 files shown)


recharts/build.boot
```diff
@@ -28,6 +28,6 @@
               :requires ["react" "react-dom" "prop-types"]
               :global-exports '{recharts Recharts})
    (sift :include #{#"^cljsjs" #"^deps\.cljs$"})
+   (validate-checksums)
    (pom)
-   (jar)
-   (validate)))
+   (jar)))
```
Deraen: Please keep the `validate` task present. It will perform `validate-checksum` but also other useful checks.

PR #1772, opened 2018-12-30
[reactstrap] Bumped to 7.0.0 and expanded externs

vincent-dm:
I updated the Reactstrap package to the new `7.0.0` release and also significantly expanded the externs file because a lot of classes were missing.

I tried running it locally using the boot script, but I couldn't get it working (the instructions about the script in the update scenario are very vague). But I did verify the `.js` and `.min.js` download links and calculated the MD5 checksums consistently with the previous ones.


Deraen:
Hmm, there is no need to use the scripts to test build. Just running `boot package install` as mentioned in https://github.com/cljsjs/packages/wiki/Creating-Packages#7-install-your-package-locally-and-try-it-out is enough to check that the package builds okay.

PR #1779, opened 2019-01-14
Highcharts7

johnmn3:


Deraen:
@johnmn3 Did you install this locally and test this in some project?

johnmn3:
> @johnmn3 Did you install this locally and test this in some project?

@Deraen I did. As with the prior versions, this does not include externs for the modules explicitly, but when I run the networkgraph module in advanced mode, it just works.

PR #1787, opened 2019-02-01
[Emotion] Add package v10.0.6-0 & [react-select] Upgrade Version to 2.3.0

xanatos1992:


xanatos1992:
Only failing because cljsjs.emotion is not available.

PR #1803, opened 2019-02-13
Add sax

Deraen:


bendlas:
Thanks for getting this started!

Right off the bat, I think sax-js' LICENSE requires a copy of itself included in the `jar` file https://github.com/isaacs/sax-js/blob/5aee2163d55cff24b817bbf550bac44841f9df45/LICENSE#L6

Deraen:
Good catch. (I think nearly package licenses probably require the license text to be included, but I think this will be first package to include license file)

sax/build.boot
```diff
@@ -0,0 +1,31 @@
...
+(deftask package []
+  (comp
+   (download :url (format "https://unpkg.com/sax@%s/lib/sax.js" +lib-version+))
+   (download :url (format "https://unpkg.com/sax@%s/LICENSE" +lib-version+))
+   (sift :move {#"^sax.js"  "cljsjs/sax/development/sax.inc.js"
+                #"^LICENSE" "cljsjs/sax/common/LICENSE"})
+   (minify :in "cljsjs/sax/development/sax.inc.js"
+           :out "cljsjs/sax/production/sax.min.inc.js")
+   (deps-cljs :provides ["sax" "cljsjs.sax"]
+              :global-exports '{sax sax})
```
bendlas: I think there needs to be another global export `cljsjs.sax sax`. Right now, `(require 'cljsjs.sax)` fails.

Deraen: `require cljsjs.sax` doesn't require global-exports. Using `:as sax` needs global-exports. `cljsjs.*` name exists only because those were used before global-exports was added, but in a new package it could be left out.

bendlas: Sorry, I'm a bit confused ..
So `(:require [cljsjs.sax])` is still supported, right?
If it is, then how is the Clojurescript compiler going to map `'cljsjs.sax` to `js/sax`?

bendlas: I'm on cljs `1.10.516`. `(:require [sax])` works. `cljsjs.sax` doesn't. Fighweel and nodejs.

bendlas:
Apart from the importing issue, LGTM

sax/build.boot
Deraen: I tested this locally and `(:require cljsjs.sax)` and `js/sax.parser` works in browser.

With `cljsjs.sax` Cljs doesn't map the name to `js/sax` automatically. This name exists just because the older packages use this naming convetion, but it is pretty much unnecessary (and confusing...) for new packages.

I don't exactly remember what you should do with Nodejs, perhaps `require("sax")`, though in that case your are accessing node module and not the foreign lib. But Cljsjs doesn't make much sense with Nodejs anyway.

With Nodejs target, if you do `:require [sax]` Cljs can automatically use the node module instead of Cljsjs foreign lib. (`:npm-deps` option might need to be enabled). This obviously won't work with `cljsjs.sax` name, and this is why new Cljsjs packages match the names to npm packages.

bendlas: Yeah, I was testing `(:require [sax :as s])` and `(:require [cljsjs.sax :as s])` with `(s/parser ..)` respectively. I don't get why we support the the former but not the latter, but still provide both namespaces.

> With cljsjs.sax Cljs doesn't map the name to js/sax automatically.

With `sax` it doesn't either, that's what `:global-exports` is for, right?

I guess also `:npm-deps`, as you said, but I'm not using that any time soon.

> I don't exactly remember what you should do with Nodejs, perhaps require("sax"), though in that case your are accessing node module and not the foreign lib. But Cljsjs doesn't make much sense with Nodejs anyway.

The foreign lib works just fine and it makes sense for portable code, like I'm doing with data.xml. The only requirement, is that the foreign lib doesn't access `window` or something else that's browser-specific. Btw, I already tested your packaging with node.

So .. I don't quite get it, but I don't feel like i need to get it either. Thanks for putting this into cljsjs!

PR #1804, opened 2019-02-15
Initial import of react-test-renderer.

mikesperber:
What the commit says. I picked the version to match react-renderer-shallow.  If there's interest, I'll update both in a later commit.


mikesperber:
Many thanks!

PR #1811, opened 2019-02-24
[hls] Updates hls package to latest 0.12.2

metacritical:
The latest version is 0.12.2 https://unpkg.com/hls.js@0.12.2/dist/hls.js

* Adds a boot-cljsjs-checksums.edn
* Install and builds correctly as local target.
* Updates externs file.


metacritical:
Force updates a previously added checksum file as the previous checksums was non capitalised.

PR #1815, opened 2019-02-25
[BABYLON] Fixes and Updates Baylon.js to latest.

metacritical:
* Fixes the current CDN url and Updates to latest stable Release 3.3.0
* Adds updated externs file.
* Adds checksum.edn
* Updates Build.boot to verify-checksums.


babylon/build.boot
```diff
@@ -17,11 +17,12 @@
 
 (deftask package []
   (comp
-    (download :url "http://cdn.babylonjs.com/2-2/babylon.js")
```
Deraen: Could you try

```
(comp
  (download :url (format "https://raw.githubusercontent.com/BabylonJS/Babylon.js/v%s/dist/babylon.js" +lib-version+))
  (sift :move {#"babylon.js" "cljsjs/babylon/production/babylon.min.inc.js"})
  ...)
```

So we get versioned url. Also the file doesn't need to downloaded second time, no `.min.inc.js` file is needed as the same file can be used for both development and production builds.

metacritical: 1. Sure, will do!
2. Yes that is redundant.

metacritical: Rebased and force pushed to a single commit.

Deraen:
Thanks. Note, for future packages, I recommend using `validate` task instead of `validate-checksums`: https://github.com/cljsjs/packages/commit/b053e9e114c9cbf87b7a426f1d630761265d95a6

PR #1826, opened 2019-03-02
[react-day-picker] Fix Exception in dev-mode

2knu:

### DIFF
--- a/react-day-picker/README.md
+++ b/react-day-picker/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-day-picker "7.3.0-0"] ;; latest release
+[cljsjs/react-day-picker "7.3.0-1"] ;; latest release
 ```
 [](/dependency)
 
--- a/react-day-picker/boot-cljsjs-checksums.edn
+++ b/react-day-picker/boot-cljsjs-checksums.edn
@@ -1,0 +1,2 @@
+{"cljsjs/react-day-picker/development/react-day-picker.inc.js"
+ "84E031CF3FAD5116E11A68DE484EA128"}
--- a/react-day-picker/build.boot
+++ b/react-day-picker/build.boot
@@ -1,10 +1,10 @@
 (def +lib-version+ "7.3.0")
-(def +version+ (str +lib-version+ "-0"))
+(def +version+ (str +lib-version+ "-1"))
 
 (set-env!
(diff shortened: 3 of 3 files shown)


2knu:
- You know this is only a diff to the current file in the master branch (deps-cljs :name and (pom) (jar) exists already).

-  I prefer the str function instead of format.

-  sift move I used, because it makes also an copy of the file, which is needed for this fix (the daypicker.js don't work). With your way I have to download the same file second times.

- I will push an update with validate-checksums instead of static checksum

In generally, when I'm updating an package, I only update the externs and the versions (and if needed the required dependencies). When the old build.boot (tasks) works fine, there is no reason for me to update it. Here I focused to fix the dev exception and it works fine.

metacritical:
> * You know this is only a diff to the current file in the master branch (deps-cljs :name and (pom) (jar) exists already).
Aah totally missed that.
> * sift move I used, because it makes also an copy of the file, which is needed for this fix (the daypicker.js don't work). With your way I have to download the same file second times.
That is interesting.
> * I will push an update with validate-checksums instead of static checksum
Great! 
> In generally, when I'm updating an package, I only update the externs and the versions (and if needed the required dependencies). When the old build.boot (tasks) works fine, there is no reason for me to update it. Here I focused to fix the dev exception and it works fine.
Yeah we all have different styles but this in fine thanks!

2knu:
For me its not important if there is an commit more or less (I even think that makes git history more comprehensible), but explicity in this case there are two different things: First the fix of the Bug, second a change of the checksum stuff. Therefore I think two commits are okay regardless of my preference.

But yes, generally one commit should be enought normally to create/update or fix some lib in this repo.

metacritical:
> For me its not important if there is an commit more or less (I even think that makes git history more comprehensible), but explicity in this case there are two different things: First the fix of the Bug, second a change of the checksum stuff. Therefore I think two commits are okay regardless of my preference.
> 
> But yes, generally one commit should be enought normally to create/update or fix some lib in this repo.

I agree at the end of the day it is upto the maintainer, how they accept patches.

Deraen:
It is great if contributors can update packages to use latest features in boot-cljsjs (like `validate` or `validate-checksums`) and clean the `build.boot` if they notice something that can be done better. But because the high volume of Cljsjs pull requests, I don't usually have time suggest those changes and I will just accept updates and new packages if I think they will work.

PR #1828, opened 2019-03-05
Upgrade antd to 3.14.1

sandhu:
Checklist has been followed.


Deraen:
> `(validate)` is being deprecated if you use `(validate-checksums)` that will be great!

`(validate)` is the recommended one. It calls `validate-checksums` but also other useful checks: https://github.com/cljsjs/boot-cljsjs/blob/master/src/cljsjs/boot_cljsjs/packaging.clj#L382

metacritical:
Ohh Thanks, my bad!

PR #1830, opened 2019-03-08
[vega] update vega version to 5.1.0-0

metasoarous:


metasoarous:
I seem to remember some problems last time with the way comments at the end of the generated externs file, but wasn't sure if it really ended up breaking anything or not. Please let me know if this ends up being an issue again this time around, and I'll fix all the PRs.

Thanks in advance!

PR #1833, opened 2019-03-10
[react-quill] Update to 1.1.0

DaveWM:
I've updated the version of `react-quill` from `1.0.0-beta-5` to `1.1.0`. I also bumped the versions of the `quill` and `react` deps to match.


DaveWM:
thanks @Deraen 👍

PR #1836, opened 2019-03-15
[vis] add missing network icons

smee:


vis/build.boot
```diff
@@ -27,7 +27,11 @@
                          #"^vis(.*)/dist/vis.min.js"
                          "cljsjs/vis/production/vis.min.inc.js"
                          #"^vis(.*)/dist/vis.min.css"
-                         "cljsjs/vis/production/vis.inc.css"})
+                         "cljsjs/vis/production/vis.inc.css"
+                         #"^vis(.*)/dist/img"
+                         "cljsjs/vis/production/img"
```
Deraen: Just move the files to one target. I think `production` is good choice.

In fact currently the files are only being moved to the first location (currently `production` folder because a map with less than 8 entries keeps the original order), because after first move operation the files don't exists, and the second move operation doesn't match any files. You can check this locally with `boot package target` and looking into the `target` directory.

smee:
You are right, my mistake. I updated the commit.

PR #1883, opened 2019-05-30
Material ui Updated to 4.0.2+ Material UI Icons: Updated to 4.0.1

2knu:


material-ui/build.boot
```diff
@@ -19,33 +19,14 @@
...
-                                                "@material-ui/core" MaterialUI
-                                                "@material-ui/core/styles" MaterialUIStyles
-                                                "@material-ui/core/colors" MaterialUIColors
-                                                ;; old names
-                                                "material-ui/styles" MaterialUIStyles
-                                                "material-ui/colors" MaterialUIColors
-                                                "material-ui" MaterialUI}
-                              :requires ["react" "react-dom"]}]
-              :externs [#"material-ui.ext.js"])
+   (deps-cljs :name "cljsjs.material-ui"
```
Deraen: This new deps-cljs is incorrect and will break existing apps. Please keep the old definition with provides and global-exports which match the npm names.

Deraen: You can see example of using Cljsjs material-ui with global-exports here: https://github.com/reagent-project/reagent/blob/master/examples/material-ui/src/example/core.cljs

And you can also use the Reagent example project to validate the new material-ui package works.

material-ui-icons/build.boot
```diff
@@ -38,4 +38,4 @@
               :externs [#"material-ui-icons.ext.js"])
    (pom)
    (jar)
-   (validate)))
+   (validate-checksums)))
```
Deraen: `validate` is preferred to just `validate-checksums`

material-ui/build.boot
2knu: Thanks for feedback, will update it. The main idea was to go away from npm (cause on win10 boot dont work with npm and the manual way always generates wrong checksums) and bring the build.boot in general more to the "clean" way with resolving duplicated exports (as most of other packages here did). (If you have a tip how to bring npm to work with boot on win10, I would be happy - problem is primary, that its not possible to call npm with sh on win - java.io.IOException: Cannot run program "npm")

But I can understand your effort to support older projects which uses the other kinds of exports. So I will bring it to work in the old way today :)

For testing I have a whole electron application, which is based on material-ui and is used daily in office. There I test most of the deps which I update and I will test the global exports there. Material-ui 4.0.1 works there as expected, after take a look on https://material-ui.com/guides/migration-v3/. With 4.0.0 you have for example to use `[(reagent/adapt-react-class <material-ui-component>)]` instead of `[:> <material-ui-component>]` (So your reagent example dont work anymore with 4.0.0+).

material-ui-icons/build.boot
2knu: oh okay, thought that validate-checksums is better to use. Will do it with validate from now :)

2knu:
Did requested Changes and updated to newer version 4.0.2 (material ui)

2knu:
Will split it into two different pull requests

PR #1896, opened 2019-06-12
[material-ui] Updated to 4.1.0

2knu:

### DIFF
--- a/material-ui/README.md
+++ b/material-ui/README.md
@@ -5,7 +5,7 @@ https://material-ui.com/
 
 [](dependency)
 ```clojure
-[cljsjs/material-ui "4.0.0-0"] ;; latest release
+[cljsjs/material-ui "4.1.0-0"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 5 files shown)


Deraen:
Testing with Reagent Material-UI example, neither 4.0.0 or this branch works for me:

image

2knu:
Fixed the Reagent Material-UI example here: https://github.com/reagent-project/reagent/pull/432
It's not a problem of material-ui rather than with the dependencies. react/react-dom 16.8.3 is required.

Deraen:
Oh, should have thought to check that. Thanks for looking into and fixing the reagent example.

Sorry for taking so long to merge MaterialUI update. I being more careful with this package as the packaging is unfortunately a bit complex and as this package has more users than most others.

2knu:
That's okay, I can understand your caution. I will look that the reagent-material-ui example runs for future updates next to my own mui-application.

Regardless, it might be good if cljsjs/react, cljsjs/react-dom and reagent could be updated to 16.8.6 before 16.9.0 will arrive (https://github.com/facebook/react/releases). Looks like you only have to trigger something for that :)

PR #1898, opened 2019-06-14
[react-bootstrap] updated to 1.0.0-beta.9

Frozenlock:


Frozenlock:
@Deraen
It looks like an unrelated package (material-ui) is causing a problem with checksums...
Any idea how to fix that?

PR #1904, opened 2019-06-26
[js-joda] add global-exports and prevent overwrite of the top level variable

henryw374:

### DIFF
--- a/js-joda/build.boot
+++ b/js-joda/build.boot
@@ -5,7 +5,7 @@
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
 (def +lib-version+ "1.10.1")
-(def +version+ (str +lib-version+ "-0"))
+(def +version+ (str +lib-version+ "-1"))
 
 (task-options!
  pom  {:project     'cljsjs/js-joda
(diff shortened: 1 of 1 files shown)


Deraen:
I don't understand why the workaround to prevent overwrite of the var is necessary. Is it something special with this library, or a general problem?

henryw374:
It's part and part. This library has addon libraries, js-joda-timezone etc, which when loaded, modify the global js-joda object. 

It turns out that on node, doing a require of a foreign-lib causes the js code to be re-run. what that means in practice is that in a project with multiple js-joda requires (from different nss), the global js-joda object is reassigned multiple times and different parts of the project referring to different instances of the global js-joda object. when the addons are loaded, on the most recent is modified. 

So, I realise that foreign-libs is not really meant for nodejs, but being that this lib has no dependencies of it's own, I think it would work well enough in this case. IOW - it doesn't hurt to do this and does prevent me getting issues raised by people who just want to quickly try things out on node and don't necessarily realise that there's a foreign lib underneath. 

Sorry that turned into a bit of a long answer!

Deraen:
Did you end up solving this in some other way?

I'd probably just recommend documenting that Node users should `npm install` the library themselves. ClojureScript (and I think Shadow-Cljs) can also automatically install npm packages based on deps.cljs on a library, like here: https://github.com/reagent-project/reagent/blob/master/src/deps.cljs

ClojureScript compiler requires `:install-deps true` option (and `:target :nodejs` to emit correct `require` call based on Cljs namespace require form). I think Shadow-CLJS installs these by default.

henryw374:
Hi, thanks for picking this up. I ended up depending on my own version of this lib that includes this change. This cljsjs package is used by my other time libraries and I doubt anyone is using it directly. I realise this change is non-standard but seems like the best compromise for now. 

I have now made it clearer to users that npm is preferable, including links to docs on how to exclude the cljsjs dependencies etc: https://github.com/juxt/tick/blob/master/docs/cljs.adoc 

So, I guess I'll close this now and revert to cljsjs dep if the situation can be resolved in any better way.

Deraen:
I think packaging your own version if perfectly fine if you need some special changes.

PR #1905, opened 2019-06-26
[react-toastify] created with version 5.3.1

2knu:


2knu:
Could be that there is something wrong with publishing the jar on maven/clojars? Cant find the artifact.

Issue #409, opened 2016-02-11
Add https://github.com/js-cookie/js-cookie


aadrian:
Please add https://github.com/js-cookie/js-cookie

Thank you.

ducky427:
@aadrian, just FYI, have you have considered using the bundled [goog.net.Cookies](https://closure-library.googlecode.com/git-history/docs/class_goog_net_Cookies.html)?

aadrian:
@ducky427 of course :). But if existing projects are already using the other API, there's not always the possibility (or the budget yet) to rewrite stuff :).

ducky427:
agreed. that makes sense.

FagnerMartinsBrack:
FWIW,  even if the application is new, generally there is no benefit in changing the module for the other one. js-cookie is simpler,  does not limit characters (accepts anything in the utf-8 range, by default using rfc 6265 for encoding), is probably smaller, have a better documentation to prevent abstraction leaks (even for server-side)  and allows converting the values into another representation to match server side incompatibilities (among other things)

Deraen:
While other points are probably true, I doubt js-cookie is smaller.

In general Closure libraries result in smaller result because Closure compiler can do dead code elimination and inlining etc. for them. Such whole program optimizations are not possible with foreign libs (cljsjs).

FagnerMartinsBrack:
Could you point a link where net.Cookies minified and gzipped is less than 900 bytes?

martinklepsch:
@FagnerMartinsBrack it doesn't matter if it's less than 900 bytes, it's exactly the size of the stuff you'll use and compressed in the most advanced way possible.

Just taking the `goog.net.Cookies` namespace, stripping comments and removing whitespace, gzipping gets me to 1084 bytes. with all names shortened by advanced compilation you're easily below 900 bytes.

Deraen:
```clj
(ns cookie-test.core
  (:require [goog.net.cookies :refer [cookies]]))

(.set cookies "name" "value")

(js/console.log (.get cookies "name"))
```

And the resulting advanced compiled file is 1.9KB, gziped 1.2KB. But that file contains general ClojureScript stuff which is present in every Cljs build. The cookie related code is something like 1.1KB and gziped 600B.

To try this yourself you can use mies template: `lein new mies cookie-test`

FagnerMartinsBrack:
@Deraen Sorry for my ignorance, but I didn't understand the code you posted. I just got really curious about the code size.

[...]

PR #1966, opened 2019-09-12
[cookie] Created with version 2.2.1

2knu:
Shortly tested it with set, get and remove a Cookie.
Releated to Request Issue https://github.com/cljsjs/packages/issues/409


2knu:
I agree and changed the package name

PR #2005, opened 2019-11-21
[React] Add parameters to componentDidCatch

aktk7z:
The new lifecycle method componentDidCatch() in React 16 receives two paramaters:
1. error
1. info

Add them to react.ext.js to match this condition.
Tested Locally.

References:

https://reactjs.org/docs/react-component.html#componentdidcatch
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L587

### DIFF
--- a/react/resources/cljsjs/react/common/react.ext.js
+++ b/react/resources/cljsjs/react/common/react.ext.js
@@ -248,7 +248,7 @@ React.Component.prototype.componentWillUnmount = function() {};
  * @param {Object} error
  * @param {Object} info
  */
-React.Component.prototype.componentDidCatch = function() {error, info};
+React.Component.prototype.componentDidCatch = function(error, info) {};
 
 /**
  * @protected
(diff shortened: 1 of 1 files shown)


react/resources/cljsjs/react/common/react.ext.js
```diff
@@ -245,8 +245,10 @@ React.Component.prototype.componentWillUnmount = function() {};
 
 /**
  * @protected
+ * @param {Object} error
+ * @param {Object} info
  */
-React.Component.prototype.componentDidCatch = function() {};
+React.Component.prototype.componentDidCatch = function() {error, info};
```
Deraen: The names are now inside the function body instead of the parameter list.

aktk7z: I've fixed the placement. Sorry for the stupid mistake.

PR #2016, opened 2019-11-30
[date-fns] Updated to 2.8.1

2knu:
Old way don't work anymore, the new one I tested locally.


date-fns/build.boot
```diff
@@ -17,12 +17,18 @@
...
+                            ["npm" "run" "build:prod"]
+                            ["rm" "-rf" "./node_modules"]])
+   (sift :move {#".*/date-fns.inc.js"  "cljsjs/date-fns/development/date-fns.inc.js"
+                #".*/date-fns.min.inc.js"  "cljsjs/date-fns/production/date-fns.min.inc.js"})
    (sift :include #{#"^cljsjs"})
-   (deps-cljs :name "cljsjs.date-fns")
+   (deps-cljs :foreign-libs [{:file           #"date-fns.inc.js"
+                              :file-min       #"date-fns.min.inc.js"
+                              :provides       ["cljsjs.date-fns"]
+                              :global-exports '{"date-fns" DateFns}}]
```
Deraen: Global exports key should match one of the provided namespaces to be useful. Here, I think `:provides` value should be `["date-fns" "cljsjs.date-fns"]`. So users can use e.g. `(:require [date-fns :refer [addDays]])`.

Name for global-exports ns should match npm name and `cljsjs.*` name can be included for compatibility for old users.

2knu: Oh okay, good point, thanks. Did it with the last force-push.

PR #2027, opened 2019-12-30
[react-stripe-elements] Add injected props in StripeProvider

danielcompton:
Stripe Elements injects some props dynamically in a way that externs generation can't detect. We added them manually to the main externs file. Is there a better way to do this instead?


Deraen:
No better way when using externs. Using externs inference for these props, or even stopping using externs and using inference for everything can be good choice though.

PR #2047, opened 2020-02-05
[blockly] add blockly

mtravers:
Adds support for Blockly, Google's Scratch-like block UI library


mtravers:
@Deraen Hi, first time doing a PR for packages...is there someone I need to ping to get this reviewed? Thanks!

Deraen:
Hi, thanks for the contribution! I'm just reviewing these quite infrequently.

This seems fine, except for those media files you included in the repo. As they are included in the archive, you should just use `sift :move` to move them to the correct path instead of including them here. Something like `#"^blockly-.*/media/(.*)" "cljsjs/blockly/common/$1"` should do the trick.

You can also support all locales, instead of just english. To move files:

`#"^blockly-.*/msg/js/(.*)\.js"             "cljsjs/blockly/development/msg/$1.inc.js"`

and to generate deps.edn entries:

```
;; Each matched file will create foreign lib entry
{:file #"cljsjs/blockly/development/msg/(.*)\.inc\.js"
 :requires ["cljsjs.blockly"]
 :provides ["cljsjs.blockly.msg.%1$s"]}
```

And to if you want to make it possible for users to select if they use the package from cljsjs or npm (with shadow-cljs or such), you should also match npm require names for provides names. Good example is the highlight package:
https://github.com/cljsjs/packages/blob/master/highlight/build.boot

mtravers:
Good suggestions, thanks! I implemented the first two, but I don't know much about npm so leaving that as is for now.

mtravers:
@Deraen thanks!

PR #2055, opened 2020-03-01
[ExcelJS] Updated to 3.8.0

2knu:


exceljs/build.boot
```diff
@@ -17,18 +17,14 @@
...
+   (download
+    :url (format "https://unpkg.com/exceljs@%s/dist/exceljs.min.js" +lib-version+)
+    :target "cljsjs/exceljs/production/exceljs.min.inc.js")
    (sift :include #{#"^cljsjs"})
-   (deps-cljs :foreign-libs [{:file #"exceljs.inc.js"
-                              :file-min #"exceljs.min.inc.js"
-                              :provides ["exceljs" "cljsjs.exceljs"]
-                              :global-exports '{"exceljs" EXCELJS}}]
-              :externs [#"exceljs.ext.js"])
+   (deps-cljs :name "cljsjs.exceljs")
```
Deraen: Since the previous version provided `exceljs` and `global-exports`, it would be good idea to provide them in new version also.

2knu: Did it in Force push. Thought that it is not that important on this lib, but its okay to keep the provided names. Global Export didn't work without foreign libs (boot-cljsjs throws an java.lang.IllegalArgumentException: option :global-exports must be of type {sym sym} - But i just copied it from old build.boot) ?

Deraen: Yes, it is probable no one is using the global export, but there is no way to know so it is best to keep the option.

You probably tried to use global-exports task option, which is very limited. You should instead use `foreign-libs` and `externs` task options, where `foreign-libs` is map with all the options, like the old version.

2knu: Ah okay, thanks for Feedback, good to know :) changed it back to the old deps-cljs task and tested it in one of my projects

PR #2060, opened 2020-04-09
[snabbdom] create package for version 0.7.4

mitchdzugan:
sample code in README was tested with advanced compilation


Deraen:
So all these modules are used alone, i.e. they don't require each other or the main module?

I'd recommend naming the foreign-libs same as node modules, `snabbdom/modules/class` instead of `snabbdom/modules/class` and adding `global-exports` definitions `:global-exports {"snabbdom/modules/class" "snabbdom_style"}`. Then users can use `(:require ["snabbdom/modules/class" :as sclass])` etc. `deps-cljs` task can be useful for creating the deps file: https://github.com/cljsjs/packages/blob/master/react-test-renderer/build.boot#L30-L40

(And depending on your project, you might want to use [Shadow-CLJS](https://shadow-cljs.github.io/docs/UsersGuide.html) or [Webpack](https://clojurescript.org/guides/webpack), instead of packaging JS libs in Cljsjs)

mitchdzugan:
The modules don't explicitly depend on the main module. They are basically useless outside of supplying them to the `init` function from the main module but they do not have any runtime dependencies on it so I figured it was ok to leave them without the `requires` key.

I updated with these commits: https://github.com/cljsjs/packages/pull/2060/files/98dde3ca8946c79e62aebd631d861587e9b34f2e..92b3edce3a24196b807138511e3d89c448b219b6

I tried using the module names like `snabbdom/modules/class` but compiled complained when I tried to require them due to invalid symbol (I assume because multiple "/"s). So while I switched to the npm module structure, I kept the "."s as my separators but if I am missing something would be happy to change.

I also wasn't sure where it made sense to use the `cljsjs.` prefix so I went ahead and made everything work either with or without it but if that is bad practice let me know.

Thanks!

Deraen:
When you use names with `/` in them, you need to make the name in the require form a string, and you need recent enough Cljs compiler (1.10.439+ probably): https://github.com/reagent-project/reagent/blob/master/examples/material-ui/src/example/core.cljs#L5

cljsjs.* names don't really make sense with global-exports. Names in `global-exports` should match exactly what is used in node `require` calls, or the code only works with cljsjs libs.

(You might also want to add `deps.cljs` to your library, as shadow-cljs users are unable to use Cljsjs libs: https://github.com/reagent-project/reagent/blob/master/src/deps.cljs. The next ClojureScript release makes it also quite easy (easier than currently) to consume libraries using libs from npm with webpack).

mitchdzugan:
Ok cool that all makes sense. I'm still pretty new to clojurescript so I appreciate all of the explanations! I have updated to use the npm paths, got rid of the nonsensical global externs, updated the README to reflect the usage with the npm paths and tested that code with advanced compilation turned on.

PR #2066, opened 2020-04-23
[snabbdom] fix deps.cljs bad paths

mitchdzugan:
Noticed this while testing with shadow-cljs. Tested the corrected version after `boot package install target` locally with shadow-cljs and works.


Deraen:
Please also update cljsjs version number so a new package gets deployed:

PR #2068, opened 2020-04-24
latest versions of js-joda core, locale-en-us and timezone. now with …

henryw374:
…global exports that match npm package names

### DIFF
--- a/js-joda-timezone/build.boot
+++ b/js-joda-timezone/build.boot
@@ -3,7 +3,7 @@
 
 (require '[cljsjs.boot-cljsjs.packaging :refer :all])
 
-(def +lib-version+ "2.0.1") 
+(def +lib-version+ "2.2.0") 
 (def +version+ (str +lib-version+ "-0"))
 
 (task-options!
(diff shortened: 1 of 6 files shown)


js-joda-timezone/build.boot
```diff
@@ -16,13 +16,16 @@
...
-    (sift :move {#"^js-joda-timezone-([\d\.]+)/dist/js-joda-timezone\.js$" "cljsjs/js-joda-timezone/development/js-joda-timezone.inc.js"})
+    (sift :move {#"^js-joda--js-joda-timezone-([\d\.]+)/packages/timezone/dist/js-joda-timezone\.js$" "cljsjs/js-joda-timezone/development/js-joda-timezone.inc.js"})
     (minify :in  "cljsjs/js-joda-timezone/development/js-joda-timezone.inc.js"       
             :out "cljsjs/js-joda-timezone/production/js-joda-timezone.min.inc.js")
     (sift :include #{#"^cljsjs"})
-    (deps-cljs :no-externs true :name "cljsjs.js-joda-timezone" :requires ["cljsjs.js-joda"] )
+    (deps-cljs 
+      :no-externs true 
+      :provides ["@js-joda/timezone"]
+      :requires ["@js-joda/core"])
```
Deraen, line 29: Timezone doesn't export a object so global-exports isn't needeed?

henryw374: yes only @js-joda/core adds global-exports. timezone and locale just do :provides here

henryw374: sorry I should explain how it works, when timezone executes, it modifies the global JSJoda object - extending its capabilities

PR #2088, opened 2020-05-15
Merge pull request #1 from csnorton/update-material-ui-chip-input-ver…

csnorton:
…sions

Removing AutoComplete dependency and updating versions.  Also refactored configs to updated syntax and frameworks.


Deraen:
There doesn't seem to be any commits here? And the autocomplete dependency was already removed in the previous update.

PR #2098, opened 2020-06-19
[react-flip-move] Update to 3.0.4

rm-hull:

### DIFF
--- a/react-flip-move/README.md
+++ b/react-flip-move/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-flip-move "3.0.1-1"] ;; latest release
+[cljsjs/react-flip-move "3.0.4-2"] ;; latest release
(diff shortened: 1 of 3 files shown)


rm-hull:
hi @Deraen whats the normal turnaround time for merging PRs? I know its only been an hour since I created the PR, but I discovered a bug in our client'a code which uses v3.0.1 of react-flip-move (but is fixed in v3.0.4). 

I just need to mitigate the situation with the client and possibly look at another solution if you cant merge.

Regards
Richard

Deraen:
Turnaround is varies from 5 minutes to 2 weeks.

If you depend on several JS packages in addition to React, I'd recommend looking into Shadow-cljs and Cljs bundle target https://clojurescript.org/news/2020-04-24-bundle-target as alternatives to Cljsjs. I'll add message to Cljsjs repository at some point recommending those.

PR #2120, opened 2020-07-22
[complex] Add support for complex.js package

sritchie:
This PR adds the https://github.com/infusion/Complex.js/ library for working with complex numbers.


sritchie:
@Deraen just wanted to check in and see if there is anything more I can do on this batch of PRs! It would be wonderful to get these libraries in. In the meantime, I'm of course happy to namespace and push them to clojars, but figured I'd ping.

Thanks!

sritchie:
By the way, I wrote up a guide for how to publish a dependency to Clojars if you need it a little faster than the official release project: https://samritchie.io/publishing-cljsjs-jars-to-clojars/

If y'all want to take this for documentation somewhere I think it might be useful for others.

Thanks!

sritchie:
@Deraen , just wanted to ping!

Deraen:
Hi,

I was on vacation. Merged now.

Packaging JS libraries yourself is perfectly fine, and I'd even recommend that in some cases. Though I'd use lein or clj instead of boot now.

Cljsjs is probably going to be at least partially deprecated at some point, as there is better ways to consume JS libraries now, drafting notice here: https://github.com/cljsjs/packages/issues/2127

PR #2129, opened 2020-08-22
[bigfraction] Add support for the bigfraction.js code inside the Fraction.js package

sritchie:
This PR adds support for the bigfraction.js version of the https://github.com/infusion/Fraction.js library, for working with rational numbers.


Deraen:
I've updated the CI image and looks like using more recent Node fixed the problem.

sritchie:
Awesome, thank you @Deraen !

Issue #2127, opened 2020-08-17
Partial deprecation of Cljsjs


Deraen:
I'm thinking Cljsjs is starting to get mostly unnecessary now, with better options for consuming JS modules available.

## TLDR

The most common packages like React will continue to be maintained in Cljsjs. But there are better alternatives for using  JS libraries.

## Why does Cljsjs exist

- To allow easy use of JS libraries from Cljs -> **No longer needed**
- To ensure there is only single copy of the JS library in classpath -> **Still relevant, in some cases**

## Why isn't Cljsjs perfect

- Centralized repository needs maintainers to review and merge PRs
- Outdated tooling, open issues with tooling

### Using a JS library from a application project

Use [shadow-cljs](https://shadow-cljs.github.io/docs/UsersGuide.html) or [ClojureScript with Webpack](https://clojurescript.org/guides/webpack) to directly consume npm packages. You always get access to up-to-date versions.

### Using a random JS library from a Cljs library

Depending on your library, you could just require that the consumers of your library use shadow-cljs or webpack to provide requires JS libraries. If you need to ensure users can use the library with foreign-libraries, you could package the JS library yourself as a JAR, same as what cljsjs does, but you can do this yourself using Leiningen or clj.

Compared to Cljsjs, you need to create a bit of tooling yourself, but you are free to update the JS library yourself.

### Using a common JS library from a Cljs library

If the JS library you are using is likely to be used by multiple Cljs libraries, using Cljsjs is useful as this ensures only one copy of the JS library is included in the build.

E.g. If multiple Cljs libraries used their own React JAR packages, multiple copies of React would get loaded.

Folcon:
I'd like to pipe up and say that at least from my perspective the below statement is not always true =)...
- To allow easy use of JS libraries from Cljs -> No longer needed

From my perspective the process of adding even one npm lib to a project can be incredibly hit and miss and I try to set aside at least a few days to ensure that it's working. Thankfully once it's running then it's usually possible to just leave it be.

Contrast that with any clojure/script lib which is basically copy deps, read the docs to work out how to use it...

This may be a minority opinion, but I don't believe that requiring all cljs devs to also maintain a detailed understanding in every javascript tool that's the new way to make things run in order to use npm libs is a great thing?

cljsjs at the very least provides a fairly consistent level of approachability to npm libraries. Sure they may be a tad out of date, but that can be tackled.

[...]

PR #2148, opened 2020-10-21
[highlight] Updated to 10.3.1

2knu:

### DIFF
--- a/highlight/README.md
+++ b/highlight/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/highlight "9.12.0-2"] ;; latest release
+[cljsjs/highlight "10.3.1-0"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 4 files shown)


Deraen:
@2knu Seeing how many packages you are using, have you seen this: https://github.com/cljsjs/packages/issues/2127 ?

I also sent you an invite to the cljsjs org, so you can merge these updates yourself.

2knu:
@Deraen Oh thanks for invitation! 

Yes I see the issue, but I primary maintain the cljsjs packages for the company which I work for and push them to our private artefact repository. So I thought it might be helpful for others, when I push it here too :)
We use a lot of cljsjs packages in different projects and for us it's the easiest way for usage and maintenance to use cljsjs packages instead of npm. Also some lein plugins like ancient, deps, licences etc. dont work with npm dependencies.

Deraen:
Yeah, makes sense.

As a note, npm of course has own set of tools which provide same features. But using two sets of tools is obviously inconvenient.

https://www.npmjs.com/package/npm-check-updates
https://github.com/dylang/npm-check
https://www.npmjs.com/package/license-checker

2knu:
Thanks for tooling hints, but as you said we want to use only one set of tooling.
Also we dont want to define and delivery the externs in every project and an extra project only for npm dependencies is a little bit superfluous. So what I want to say overall: cljsjs/packages is awesome, really helpful for us and upgrading packages is mostly very easy and requires only a small effort :)

Deraen:
Externs aren't used at all with Shadow-cljs or Cljs+Webpack. I guess I could add mention about that to the draft.

PR #2151, opened 2020-10-29
[update] d3 to version 6.2

smee:


smee:
The externs seem to be insufficient, will check further

smee:
Strange: I had problems with advanced optimizations, which only went away after using the clojurescript compiler option `:infer-externs`. All my experiments trying to replicate the problem using the CLI version of the closure compiler were not successful: I could not replicate my problems.

So, if the old externs file was working, this one should work as well.

PR #2153, opened 2020-11-11
[vega] [vega-lite] [vega-embed] Update versions

mooreryan:
First time contributing to cljsjs...tried to follow all the guidelines, but let me know if there were any issues!

- Update library versions
- Switch to using `(validate-checksums)` (as prompted by the build script)


mooreryan:
Looking at the CircleCI output, looks like the error is in vega-embed depending on newer version of vega and vega-lite, bet they haven't actually been built on clojars yet (as I'm attempting to update them here).  

Not sure how to fix this :/

mooreryan:
I'm going to close this.  After reading https://github.com/cljsjs/packages/issues/2127, it's easy enough to just use shadow-cljs for this.

PR #2154, opened 2020-12-08
[alasql] create alasql package

kitrerp:
AlaSQL - is a lightweight client-side in-memory SQL database designed to work in browser and Node.js.


kitrerp:
Does this PR look good? @Deraen

PR #2205, opened 2021-08-05
[react-leaflet] Make core API accessible

bgalartza:
The version has been tested and used in production for several months, but as it's the first time creating a package of this type (using npm) a review would be appreciated. Thank you in advance.

### DIFF
--- a/react-leaflet/README.md
+++ b/react-leaflet/README.md
@@ -2,7 +2,7 @@
 
 [](dependency)
 ```clojure
-[cljsjs/react-leaflet "3.1.0-0"] ;; latest release
+[cljsjs/react-leaflet "3.1.0-1"] ;; latest release
 ```
 [](/dependency)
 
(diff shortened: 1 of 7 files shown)


Deraen:
Hmm, is there really some part of react-leaflet API that is completely inaccessible through the official UMD JS file? That's unfortunate. I'd really prefer using the existing files whenever possible.

I don't have any projects using this myself but I'm asking a co-worker if we still have some other projects using cljsjs instead of shadow-cljs.

bgalartza:
As far as we could see the [core api](https://react-leaflet.js.org/docs/v3/core-introduction/) which is used to extend the functionality of the library is not available in the main UMD JS file. But after re-reading the documentation it looks like it's distributed in a separated package(which is a dependency of the main one, but apparently not included in the main JS file).

For the new 4.x versions they have changed the way they distribute the library. We will revisit the topic once we decide to update the library in our projects and make a new PR with a detailed explanation of our findings. 

In the meantime I'm closing the issue. Thank you for your time.

